PR #103, opened 2025-09-23
Initial Postgres integration

carte7000:


aggregator/migrations/postgres/00001_create_all_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+    participant_id TEXT NOT NULL DEFAULT '',
+    signer_address TEXT NOT NULL,
+    source_chain_selector TEXT NOT NULL,
+    dest_chain_selector TEXT NOT NULL,
+    onramp_address TEXT NOT NULL,
+    offramp_address TEXT NOT NULL,
+    signature_r BYTEA NOT NULL DEFAULT '',
+    signature_s BYTEA NOT NULL DEFAULT '',
+    ccv_node_data BYTEA NOT NULL,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
```
asoliman92, line 18: I think we'll need an index on the date as well for efficient querying?

asoliman92: same for the rest of the tables

aggregator/migrations/postgres/00001_create_all_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+CREATE INDEX IF NOT EXISTS idx_commit_aggregated_reports_timestamp ON commit_aggregated_reports(timestamp);
+
+-- Create indexes for block_checkpoints
+CREATE INDEX IF NOT EXISTS idx_block_checkpoints_client_id ON block_checkpoints(client_id);
+CREATE INDEX IF NOT EXISTS idx_block_checkpoints_chain_selector ON block_checkpoints(chain_selector);
+CREATE INDEX IF NOT EXISTS idx_block_checkpoints_updated_at ON block_checkpoints(updated_at);
+
+-- +goose Down
+-- Drop indexes first
+DROP INDEX IF EXISTS idx_block_checkpoints_updated_at;
```
asoliman92, line 62: Why are we dropping after creating the indexes?

aggregator/migrations/postgres/00001_create_all_tables.sql
asoliman92: I guess it's okay then and add it if we need it later for different kind of queries

aggregator/migrations/postgres/00001_create_all_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    
+    CONSTRAINT unique_aggregated_report UNIQUE (message_id, committee_id)
+);
+
+-- Create block_checkpoints table
+CREATE TABLE IF NOT EXISTS block_checkpoints (
+    id BIGSERIAL PRIMARY KEY,
+    client_id TEXT NOT NULL,
+    chain_selector TEXT NOT NULL,
```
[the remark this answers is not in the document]
carte7000: Yes I can add it if needed. My assumption is that if the verifier is supporting multiple committee it would share the same finalized block across all of them but if we want to partition this I can add the field. Another way of doing that would be to have 2 sets of credentials but probably easier to just allow the client to send a "discriminator" key that allow them to split how they want

aggregator/migrations/postgres/00001_create_all_tables.sql
```diff
@@ -25,24 +16,20 @@ CREATE TABLE IF NOT EXISTS commit_verification_records (
     signature_s BYTEA NOT NULL DEFAULT '',
     ccv_node_data BYTEA NOT NULL,
     created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
-    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
-    
-    CONSTRAINT unique_verification UNIQUE (message_id, signer_address, committee_id)
+    CONSTRAINT unique_verification_sequence UNIQUE (message_id, committee_id, signer_address, seq_num)
```
asoliman92, line 19: don't really think you need to add the `seq_num` in the constraint. only the previous 4 are relevant and you always query for all of them and check the latest seq_num. wdyt?

aggregator/migrations/postgres/00001_create_all_tables.sql
```diff
@@ -25,24 +16,20 @@ CREATE TABLE IF NOT EXISTS commit_verification_records (
...
+    seq_num BIGINT NOT NULL DEFAULT nextval('commit_aggregated_reports_seq_num_seq'),
     message_id TEXT NOT NULL,
     committee_id TEXT NOT NULL,
+    verification_record_ids BIGINT[] NOT NULL,
     report_data BYTEA NOT NULL,
     created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
-    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
-    
-    CONSTRAINT unique_aggregated_report UNIQUE (message_id, committee_id)
+    CONSTRAINT unique_aggregated_report_sequence UNIQUE (message_id, committee_id, seq_num)
```
asoliman92, line 30: same

PR #107, opened 2025-09-24
feat: indexer config

KodeyThomas:
In this PR, the indexer configuration has been added + auto generation of the config.toml file from devenv

### DIFF
--- a/indexer/cmd/main.go
+++ b/indexer/cmd/main.go
@@ -20,6 +20,7 @@ import (
 )
 
 func main() {
+	// Load in the config from './config.toml'
 	config, err := config.LoadConfig()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


carte7000 (approved):

indexer/config.example.toml
```diff
@@ -0,0 +1,27 @@
...
+LogStreamingEnabled = true
+MetricReaderInterval = 5
+TraceSampleRatio = 1.0
+TraceBatchTimeout = 10
+
+[Scanner]
+ScanInterval = 1
+
+[Discovery]
+Type = "static"
```
winder, line 17: Why do you need the type here? The subsequent `Discovery.Static.*` headers include it already.

PR #112, opened 2025-09-24
Time to aggregate metric

carte7000:


aggregator/pkg/monitoring/metrics.go
```diff
@@ -21,6 +21,7 @@ type AggregatorMetrics struct {
 	apiRequestError                        metric.Int64Counter
 	getMessageSinceNumberOfRecordsReturned metric.Int64Histogram
 	pendingAggregationsChannelBuffer       metric.Int64UpDownCounter
+	timeToAggregation                      metric.Int64Histogram
```
KodeyThomas, line 24: as it's a histogram you'll need to create the `metricViews` for it just need to define the buckets etc

PR #113, opened 2025-09-24
Move proto to chainlink proto

carte7000:


aggregator/pkg/server.go
```diff
@@ -28,15 +28,16 @@ import (
 	"github.com/smartcontractkit/chainlink-ccv/aggregator/pkg/monitoring"
 	"github.com/smartcontractkit/chainlink-ccv/aggregator/pkg/quorum"
 	"github.com/smartcontractkit/chainlink-ccv/aggregator/pkg/storage"
-	"github.com/smartcontractkit/chainlink-ccv/common/pb/aggregator"
 	"github.com/smartcontractkit/chainlink-common/pkg/beholder"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+
+	pb "github.com/smartcontractkit/chainlink-protos/chainlink-ccv/go/v1"
```
0xAustinWang, line 34: weird that it's imported as `pb` here, but `aggregator` everywhere else. 

Also weird that we have a named import as `aggregator` inside the `aggregator` component 😆 

Can we rename the other imports to all be `pb`?

0xAustinWang (approved):

PR #116, opened 2025-09-25
Recover orphan

carte7000:


aggregator/pkg/storage/metrics_aware_storage.go
```diff
@@ -59,6 +59,10 @@ func (s *MetricsAwareStorage) ListCommitVerificationByMessageID(ctx context.Cont
 	})
 }
 
+func (s *MetricsAwareStorage) ListOrphanedMessageCommitteePairs(ctx context.Context) (<-chan *model.MessageCommitteePair, <-chan error) {
+	return s.inner.ListOrphanedMessageCommitteePairs(ctx)
```
asoliman92, line 63: why not capturing metrics here?

PR #117, opened 2025-09-25
Extend beholder config for aggregator

carte7000:


aggregator/pkg/model/configuration.go
```diff
@@ -119,6 +119,36 @@ type CheckpointConfig struct {
 	MaxCheckpointsPerRequest int `toml:"maxCheckpointsPerRequest"`
 }
 
+// BeholderConfig wraps the beholder configuration to expose a minimal config for the aggregator.
+type BeholderConfig struct {
+	// InsecureConnection disables TLS for the beholder client.
```
asoliman92, line 124: I think this will be the same for all services. Should we add it under protocol and use it by them all including devenv?

PR #118, opened 2025-09-25
configs from env vars

AndresJulia:


0xAustinWang (approved):
Can you make sure the devenv sets these on startup?

PR #119, opened 2025-09-25
Flatten verifier.

winder:
Move types and interface definitions into the root verifier package.


verifier/interfaces.go
```diff
@@ -1,12 +1,21 @@
-package reader
+package verifier
 
 import (
 	"context"
 	"math/big"
 
-	"github.com/smartcontractkit/chainlink-ccv/verifier/pkg/types"
+	protocol "github.com/smartcontractkit/chainlink-ccv/protocol/pkg/types"
```
0xAustinWang, line 7: you'll have conflicts with the "flatten protocol" pr

winder: I branched off of this one for the other PR, should be alright.

asoliman92 (approved):

PR #130, opened 2025-09-30
Use new call from ccv aggregator for getting required ccvs

0xAustinWang:
Update to use the most recent `chainlink-ccip/chains/evm/deployments` version that has a new call in ccv aggregator

This also includes the rename from on/offramp to committeeVerifier

### DIFF
--- a/build/devenv/tests/e2e/smoke_test.go
+++ b/build/devenv/tests/e2e/smoke_test.go
@@ -104,9 +104,9 @@ func TestE2ESmoke(t *testing.T) {
 			threshold       uint8
 		}
 
-		verifierAddress := common.HexToAddress("0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1")
-		execOnRamp := "0x9A9f2CCfdE556A7E9Ff0848998Aa4a0CFD8863AE"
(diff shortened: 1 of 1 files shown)


build/devenv/cmd/ccv/ccv.go
```diff
@@ -479,12 +479,12 @@ var sendCmd = &cobra.Command{
...
 
 			return impl.SendArgsV3Message(ctx, e, in.CLDF.Addresses, selectors, src, dest, uint16(finality),
-				"0x9A9f2CCfdE556A7E9Ff0848998Aa4a0CFD8863AE", // executor address
+				"0x68B1D87F95878fE05B998F19b66F4baba5De1aed", // executor onramp address
 				"0x3Aa5ebB10DC797CAC828524e59A333d0A371443c", // mock receiver
 				nil, nil,
 				[]protocol.CCV{
 					{
-						CCVAddress: common.HexToAddress("0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1").Bytes(),
+						CCVAddress: common.HexToAddress("0x0B306BF915C4d645ff596e518fAf3F9669b97016").Bytes(),
```
kylesmartin, line 487: Ideally can read from datastore at some point, so we don't have to update this in-code whenever deployment changes

PR #141, opened 2025-10-02
build/devenv: remove hardcoded addresses from test

makramkd:


cciptestinterfaces/interface.go
```diff
@@ -25,23 +25,75 @@ for CCIP16 and CCIP17
...
 }
 
-// Testable provides functions for a standardized CCIP test suite.
-type Testable interface {
-	// SendArgsV2Message sends a message with ArgsV2 params
-	SendArgsV2Message(ctx context.Context, e *deployment.Environment, addresses []string, src, dest uint64) error
-	// SendArgsV3Message sends a message with ArgsV3 params
-	SendArgsV3Message(ctx context.Context, e *deployment.Environment, addresses []string, selectors []uint64, src, dest uint64, finality uint16, execAddr, receiverAddr string, execArgs, tokenArgs []byte, ccv, optCcv []protocol.CCV, threshold uint8) error
+// TokenAmount represents a token amount being sent in a CCIP message.
+type TokenAmount struct {
```
makramkd: After adding these types implementing the interface in ccv-evm was causing circular imports. We could potentially move this to `protocol` if it makes sense to authors.

PR #149, opened 2025-10-03
Simon/aggregator/orphan recovery

carte7000:
## Overview
This PR introduces an orphan recovery mechanism to handle cases where commit verification records are successfully stored but fail to trigger aggregation due to transient errors or system failures.

## Problem Statement
When commit verification records are submitted to the aggregator, they should trigger an aggregation check to determine if enough verifications exist to meet the quorum threshold. However, if the aggregation trigger fails due to:

**Transient network errors**
Temporary service unavailability
Race conditions
System crashes
The verification records become "orphaned" - they exist in storage but were never processed for aggregation, potentially causing messages to never reach quorum despite having sufficient verifications.

## Solution Algorithm
1. Orphan Marking (During Record Creation)
When saving commit verification records in [DynamoDBStorage.SaveCommitVerification](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html):

Each accumulator record gets a sparse GSI field PendingAggregation with value "PENDING_{committeeID}"
This creates an efficient index for scanning orphaned records by committee

2. Orphan Recovery Process
The [OrphanRecoverer](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) runs periodically and:

Scan for Orphans: Query the PendingAggregation GSI to find all accumulator records still marked as pending for each committee
Extract Message IDs: From each orphaned accumulator record, extract the [MessageID](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Trigger Re-aggregation: Call [aggregator.CheckAggregation(messageID, committeeID)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to re-evaluate if the message now has sufficient verifications for quorum
Automatic Cleanup: When aggregation succeeds and a report is submitted via [SubmitReport](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), the PendingAggregation field is removed and status updated to "AGGREGATED"

3. Configuration-Driven Execution
Orphan recovery is configurable via OrphanRecovery.Enabled and OrphanRecovery.IntervalSeconds
Runs continuously in background when enabled
Graceful shutdown on context cancellation

## Key Benefits
Resilient Message Processing: Ensures no messages are permanently stuck due to transient failures
Automatic Recovery: No manual intervention required - the system self-heals
Efficient Scanning: Uses DynamoDB GSI for fast orphan detection by committee
Minimal Performance Impact: Sparse GSI only contains pending records, cleanup removes processed items

## Implementation Details
Sparse GSI field only exists for unprocessed records
Streaming Results returns channels for memory-efficient processing of large result sets
Idempotent Operations: Re-triggering aggregation is safe - already aggregated messages are skipped
Error Handling: Individual orphan processing failures don't stop the entire recovery cycle
This ensures the aggregator service maintains high reliability and eventually processes all submitted verification records, even in the face of transient system failures.


aggregator/pkg/model/configuration.go
```diff
@@ -206,6 +213,7 @@ type AggregatorConfig struct {
 	Storage           StorageConfig              `toml:"storage"`
 	APIKeys           APIKeyConfig               `toml:"apiKeys"`
 	Checkpoints       CheckpointConfig           `toml:"checkpoints"`
+	OrphanRecovery    OrphanRecoveryConfig       `toml:"orphanRecovery"`
```
asoliman92, line 216: Better get this out of BeholderConfig if possible. Already I extracted the `BeholderConfig` under `protocol/config` so that we can use the same in multiple services. Any extra/specific configs better do them as a standalone ones

aggregator/pkg/model/configuration.go
```diff
@@ -137,6 +137,13 @@ type CheckpointConfig struct {
 	MaxCheckpointsPerRequest int `toml:"maxCheckpointsPerRequest"`
 }
 
+type OrphanRecoveryConfig struct {
+	// Enabled controls whether orphan recovery is enabled
+	Enabled bool `toml:"enabled"`
```
asoliman92, line 142: In 1.6 we used to use `Disabled` as the main config. That way if someone leaves it then it defaults to being `Enabled`. That said, I don't have a bias towards smth other than consistency with what we had before but you can argue that it's a new protocol and doesn't need to follow the same way.

aggregator/pkg/storage/ddb/dynamodb_storage_test.go
```diff
@@ -391,3 +392,179 @@ func TestAggregatedReportOperations(t *testing.T) {
...
+		foundmessageIds := make(map[string]bool)
+		for _, orphan := range orphans {
+			key := fmt.Sprintf("%x", orphan)
+			foundmessageIds[key] = true
+		}
+
+		expectedKey1 := fmt.Sprintf("%x", messageID1)
+		expectedKey2 := fmt.Sprintf("%x", messageID2)
+
+		require.True(t, foundmessageIds[expectedKey1], "Should find message1 messageId")
```
asoliman92, line 468: AFAIU messag1 shouldn't be an orphan because 2 participants verified it?

aggregator/pkg/orphan_recoverer.go
```diff
@@ -0,0 +1,138 @@
...
+		err := o.RecoverOrphans(ctx)
+		if err != nil {
+			o.logger.Errorw("Orphan recovery scan failed", "error", err)
+		} else {
+			o.logger.Info("Orphan recovery scan completed successfully")
+		}
+		duration := time.Since(now)
+		o.logger.Infow("Orphan recovery scan finished",
+			"duration", duration)
+		if duration < time.Duration(orphanRecoveryConfig.IntervalSeconds)*time.Second {
```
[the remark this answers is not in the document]
carte7000: I had use the timer initially, but the goal was to avoid having multiple concurrent run that's why I added the sleep instead of a timer that way if the run last for more than the timer duration we won't have multiple run pilling up together

aggregator/pkg/storage/ddb/dynamodb_storage_test.go
carte7000: That's correct but in this particular test we never do an aggregation. The record are removed from the pending list only when we call Submit report that's why in this particular test we have many verification record without any of them being aggregated/removed from pending

aggregator/pkg/model/configuration.go
carte7000: We use Enabled in multiple diff config in 1.7 for instance we have Enabled for the beholder metrics in most service currently

aggregator/pkg/storage/ddb/dynamodb_storage.go
```diff
@@ -358,3 +403,80 @@ func (d *DynamoDBStorage) queryFinalizedFeedByGSI(ctx context.Context, gsiPK str
...
+				select {
+				case errorChan <- fmt.Errorf("failed to scan orphaned records: %w", err):
+				case <-ctx.Done():
+				}
+				return
+			}
+
+			// Process results
+			for _, item := range result.Items {
+				if !accumulatorDTO.IsAccumulatorRecord(item) {
```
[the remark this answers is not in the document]
carte7000: I renamed it verificationMessageData

PR #162, opened 2025-10-08
feat(indexer): persistent postgres backed storage

KodeyThomas:
This PR adds in a persistent storage backend using postgres and goose to manage migrations. 

Additionally, a new storage of `sink` to aggregate multiple storage backends together to allow for caching layers and writing to multiple storage backends at the same time.

### DIFF
--- a/build/devenv/services/indexer.go
+++ b/build/devenv/services/indexer.go
@@ -104,13 +104,35 @@ func defaults(in *IndexerInput) {
 				},
 			},
 			Storage: config.StorageConfig{
-				Type: "postgres",
-				Postgres: config.PostgresConfig{
-					URI:                    "postgresql://indexer:indexer@indexer-db:5432/indexer?sslmode=disable",
-					MaxOpenConnections:     20,
-					MaxIdleConnections:     5,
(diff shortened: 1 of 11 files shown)


build/devenv/services/indexer.go
```diff
@@ -131,19 +159,17 @@ func NewIndexer(in *IndexerInput) (*IndexerOutput, error) {
...
-		}
+	buff := new(bytes.Buffer)
+	encoder := toml.NewEncoder(buff)
+	encoder.Indent = ""
+	err = encoder.Encode(in.IndexerConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode config: %w", err)
+	}
+
+	err = os.WriteFile(configPath, buff.Bytes(), 0o644)
```
carte7000, line 170: Did we settle on the way of configuring the service? Are we going with static files or dynamic generation?

KodeyThomas: think having the config generated by devenv makes sense, think the discussion is still ongoing

carte7000 (approved):

PR #999, opened 2026-04-02
Add Solana chain family to enrichEnvironmentTopology

ogtownsend:
We will reuse EVM verifier signing keys for Solana messages


build/devenv/implcommon.go
```diff
@@ -240,20 +71,33 @@ func connectAllChainsLegacy(
...
 		cvInputs := buildCommitteeVerifierInputs(topology, entry.remoteSelectors, entries)
+
 		for _, rs := range entry.remoteSelectors {
+			// Normalize selector pairs to avoid duplicate lane configs for (A,B) vs (B,A).
+			lo := min(sel, rs)
+			hi := max(sel, rs)
+			if _, dup := seen[[2]uint64{lo, hi}]; dup {
+				continue
+			}
+			seen[[2]uint64{lo, hi}] = struct{}{}
```
ogtownsend: `connectAllChains` was building one `lanes.LaneConfig` per ordered chain for every remote selector, so for two chains A and B it emitted both (A, B) and (B, A). In `lanes.ConnectChains`, a single `LaneConfig` already configures the lane in both directions (source-->dest and dest-->source) and merges all resulting AddressRefs into the DS, so the second config for the same unordered pair tried to insert the same refs again and failed with “an address ref with the supplied key already exists.” The change normalizes each selector pair and dedupes. 

`lanes.ConnectChains` already treats one `LaneConfig` as a fully bidirectional lane: https://github.com/smartcontractkit/chainlink-ccip/blob/main/deployment/lanes/connect_chains.go#L175-L177

tt-cll (approved):
this is the pattern for now, so the change is fine, but maybe this chain specific registration in the centralized location is something we can abstract soon cc @0xAustinWang @winder @makramkd

ogtownsend:
@tt-cll yeah I think we need a proper registration approach for this and potentially transmitter key as well: https://github.com/smartcontractkit/chainlink-ccv/blob/main/build/devenv/services/executor/base.go#L561-L580

archseer (approved):

archseer (approved):

0xAustinWang (approved):

PR #1003, opened 2026-04-06
composable testing api

0xAustinWang:
Use the markerInterface to provide chain specific send options.
Remove chain specific functions from the generic `Chain` interface.
Create example test using smaller interfaces


build/devenv/cciptestinterfaces/interface.go
```diff
@@ -396,3 +388,27 @@ type OffChainConfigurable interface {
...
+
+type ChainAsDestination interface {
+	GetEOAReceiverAddress() (protocol.UnknownAddress, error)
+	ConfirmExecOnDest(ctx context.Context, from uint64, key MessageEventKey, timeout time.Duration) (ExecutionStateChangedEvent, error)
+
+	genericChain
+}
+
+type ChainAsSource interface {
+	BuildChainMessage(ctx context.Context, destChain uint64, messageFields MessageFields, opts MessageOptions) (any, error)
```
winder: nit: instead of using `any` here, you could have a stronger type. Even if it's just `type CASMessage any`. Return `CASMessage` here and accept it in `SendChainMessage` and the user knows exactly what they're supposed to do because the type system insists on the way it should be used.

winder (dismissed):
Looks like a nice incremental improvement.

winder (dismissed):

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -396,3 +388,27 @@ type OffChainConfigurable interface {
 	// using chain-specific clients or CLDF
 	FundAddresses(ctx context.Context, bc *blockchain.Input, addresses []protocol.UnknownAddress, nativeAmount *big.Int) error
 }
+
+type ChainSendOption interface {
+	IsSendOption() bool
+}
```
makramkd, line 397: Can we have some docstrings on this, like when it should be implemented + what it achieves.

build/devenv/cciptestinterfaces/interface.go
makramkd: Agreed, it helps connect the dots on the interface's methods.

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -396,3 +388,27 @@ type OffChainConfigurable interface {
...
+
+type genericChain interface {
+	ChainSelector() uint64
+}
+
+type ChainAsDestination interface {
+	GetEOAReceiverAddress() (protocol.UnknownAddress, error)
+	ConfirmExecOnDest(ctx context.Context, from uint64, key MessageEventKey, timeout time.Duration) (ExecutionStateChangedEvent, error)
```
makramkd, line 412: Struct docstring + method docstrings plz

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -396,3 +388,27 @@ type OffChainConfigurable interface {
...
+
+	genericChain
+}
+
+type ChainAsSource interface {
```
makramkd, line 417: Needs struct + method docstrings

build/devenv/evm/interface.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+
+	"github.com/smartcontractkit/chainlink-ccv/build/devenv/cciptestinterfaces"
+	"github.com/smartcontractkit/chainlink-ccv/protocol"
+)
+
+type EVMOptions interface {
```
makramkd, line 13: Can we have a struct doc comment on this? I presume this is an optional interface that is expected to be implemented only by EVM chains?

build/devenv/evm/interface.go
```diff
@@ -0,0 +1,39 @@
...
+type EVMSendOptions struct {
+	Nonce                        *uint64
+	Sender                       *bind.TransactOpts
+	UseTestRouter                bool
+	DisableTokenAmountValidation bool
+}
+
+type IEVMSendOptions interface {
+	cciptestinterfaces.ChainSendOption
+	WithEVMSendOptions() *EVMSendOptions
```
makramkd: Where is this method used?

build/devenv/evm/interface.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+type EVMSendOptions struct {
+	Nonce                        *uint64
+	Sender                       *bind.TransactOpts
+	UseTestRouter                bool
+	DisableTokenAmountValidation bool
+}
+
+type IEVMSendOptions interface {
```
makramkd: The `I` prefix on Go interfaces isn't standard, wondering if this interface is needed given that the WithEVMSendOptions() method doesn't seem to be used (not sure though)

build/devenv/tests/composable/messaging/evmPOC_test.go
```diff
@@ -0,0 +1,62 @@
...
+	require.True(t, destOk, "destChain does not match the chainAsDestination interface!")
+
+	require.NoError(t, BasicMessageTestScenario(ctx, t, srcChain, destChain, cciptestinterfaces.MessageFields{
+		Receiver: receiver,
+		Data:     []byte{},
+	}, cciptestinterfaces.MessageOptions{
+		Version:             2,
+		ExecutionGasLimit:   200_000,
+		OutOfOrderExecution: false,
+	}, nil))
```
makramkd, line 61: Correct me if I'm wrong: we're passing `nil` here, which we can pretty much do for all chain-agnostic test cases, because we only need the SendOption if we're doing some chain-specific mutation (in the EVM case, account nonces)?

changelog/2026-04-20_composable_chain_interfaces.md
```diff
@@ -0,0 +1,237 @@
...
+
+// ChainAsDestination is implemented by any chain that can receive CCIP messages.
+type ChainAsDestination interface {
+    GetEOAReceiverAddress() (protocol.UnknownAddress, error)
+    ConfirmExecOnDest(ctx context.Context, from uint64, key MessageEventKey, timeout time.Duration) (ExecutionStateChangedEvent, error)
+    ChainSelector() uint64
+}
+```
+
+The removed methods (`SendMessageWithNonce`, `GetUserNonce`, `GetRoundRobinUser`) were EVM-specific and have been moved to the new `evm.EVMOptions` interface.
```
makramkd: Small correction: `SendMessageWithNonce` was removed entirely, replaced by a call to GetUserNonce and then providing the options to `SendChainMessage`

build/devenv/tests/composable/messaging/messageAsSource.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-ccv/build/devenv/cciptestinterfaces"
+)
+
+func BasicMessageTestScenario(
```
makramkd, line 16: Lets add a docstring describing this test case

build/devenv/tests/composable/messaging/evmPOC_test.go
0xAustinWang: Yea exactly. The implementation can have default behavior (query for nonce) or be manipulated based on the SendOption function that is passed in. 

For the EVM side, the information is all static, but I left the functions in as examples of how we can achieve chain specific functionality. (for example if we need to query an RPC or create additional Solana accounts etc.)

winder (approved):

makramkd (approved):

PR #1007, opened 2026-04-08
Use bootstrap.JobSpec in apps and devenv.

winder:
Convert `VerifierJobSpec` and `ExecutorJobSpec` to a generic `bootstrap.JobSpec` that can be used on all app types.

This will simplify the AccessorRegistry by avoiding generics with config types and allowing shared config handling.

### DIFF
--- a/bootstrap/job.go
+++ b/bootstrap/job.go
@@ -1,0 +1,9 @@
+package bootstrap
+
+// JobSpec is the specification for a commit verifier job, pushed by JD.
+type JobSpec struct {
+	ExternalJobID string `toml:"externalJobID"`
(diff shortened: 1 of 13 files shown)


bootstrap/job.go
```diff
@@ -0,0 +1,9 @@
+package bootstrap
+
+// JobSpec is the specification for a commit verifier job, pushed by JD.
```
makramkd:
```suggestion
// JobSpec is the specification for a bootstrapped service job, pushed by JD.
```

0xAustinWang (approved):

tt-cll (approved):

PR #1020, opened 2026-04-12
fix tp devenv

stackman27:


build/devenv/implcommon.go
```diff
@@ -526,17 +527,71 @@ func ConfigureAllTokenTransfers(
...
+
+		switch family {
+		case chainsel.FamilyEVM:
+			if current.TokenPoolRef.Type == datastore.ContractType(devenvcommon.BurnMintTokenPoolType) {
+				return current, nil
+			}
+			if candidate.TokenPoolRef.Type == datastore.ContractType(devenvcommon.BurnMintTokenPoolType) {
+				return candidate, nil
+			}
+		case chainsel.FamilyCanton:
```
tt-cll: we took this offline but we shouldn't be using chain specific switches. We should fix the filtering mechanism

tt-cll (approved):

skudasov (approved):

PR #1029, opened 2026-04-16
Add curse e2e test coverage

carte7000:


build/devenv/tests/e2e/finality_reorg_curse_test.go
```diff
@@ -388,6 +390,147 @@ func TestE2EReorg(t *testing.T) {
...
+
+		require.NoError(t, srcImpl.Uncurse(ctx, [][16]byte{chainSelectorToSubject(destSelector)}))
+
+		// With the checkpoint past the message block, uncursing on-chain alone must not replay it.
+		anvilHelper.MustMine(ctx, verifier.ConfirmationDepth+5)
+		time.Sleep(10 * time.Second)
+		verifyMessageNotExists(droppedMsgID, "Dropped message should not reappear after uncurse alone")
+
+		// Each committee member has its own DB: pause it, rewind the finalized height, restart it.
+		for _, c := range committeeContainers {
```
0xAustinWang: i wonder if the logic to replay on committee verifier should be something we encapsulate into an interface somewhere, so it can be used across other tests (finality violation etc.)

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -199,6 +199,41 @@ type LombardMailboxBridgedMessageSetter interface {
...
+type SnapshotID string
+
+// ReorgableChain is optionally implemented by chain families that can
+// snapshot and revert the node state to simulate reorgs (e.g. anvil via
+// evm_snapshot / evm_revert). Keep this distinct from ProgressableChain:
+// a curse test only needs block progression, not reorg primitives.
+type ReorgableChain interface {
+	// SupportReorgs returns true iff the underlying node supports taking
+	// snapshots and reverting to them.
+	SupportReorgs(ctx context.Context) bool
```
0xAustinWang, line 227: We don't need `SupportReorgs` or `SupportManualBlockProgress`, a chain family will either implement it or not, we get these for free by trying to cast the interface

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -199,6 +199,41 @@ type LombardMailboxBridgedMessageSetter interface {
...
+	// supports forced block progression AND is configured to mine on each
+	// tx (so messages sent by tests actually get included).
+	SupportManualBlockProgress(ctx context.Context) bool
+	// AdvanceBlocks mines numBlocks blocks on the chain.
+	AdvanceBlocks(ctx context.Context, numBlocks int) error
+}
+
+// SnapshotID identifies a snapshot taken on a ReorgableChain. It is
+// opaque to the caller; each chain family defines its own encoding.
+type SnapshotID string
```
0xAustinWang: maybe using []bytes here is better?

build/devenv/evm/block_progression.go
```diff
@@ -0,0 +1,99 @@
...
+		if err := m.ethClient.Client().CallContext(ctx, &result, "evm_mine"); err != nil {
+			return fmt.Errorf("advance blocks: evm_mine on block %d/%d: %w", i+1, numBlocks, err)
+		}
+	}
+	m.logger.Debug().Int("numBlocks", numBlocks).Msg("Advanced blocks")
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-time.After(postMineSettle):
```
0xAustinWang: postminesettle can be a parameter

build/devenv/cciptestinterfaces/interface.go
carte7000: Even within a chain family it might be supported depending on the actual RPC impl you use. For instance if you don't use anvil you might lose some of the block progress control

asoliman92 (dismissed):

build/devenv/tests/e2e/smoke_replay_cli_test.go
```diff
@@ -27,6 +28,12 @@ import (
 
 const replayBinary = "/bin/indexer-replay"
 
+func execInContainer(containerName string, args ...string) (string, error) {
```
KodeyThomas: we probably want a context here right? otherwise this could hang and become flakey

KodeyThomas (approved):

asoliman92 (approved):

PR #1031, opened 2026-04-17
offchain tooling api

tt-cll:


evm/go.mod
winder, line 1: I'm a little hesitant to bury modules in `build/devenv`. Should we put this in a new top-level package called `platform` and treat that directory as a sort of mono-repo?

tt-cll: This is the EVM impl of offchain tooling and it is at the top level :-). Don't know if there's a better home for that

Where should the EVM impl live vs the interface declarations? We can rename `chainlink-ccv/deployment` to `chainlink-ccv/platform` if that's clearer

carte7000 (dismissed):

evm/go.mod
makramkd: I think this is fine for now, maybe we can have an `evm/deployment` which has the interface implementations of the offchain deployment interfaces?

I'm imagining that in the future `evm` could also hold the evm-specific implementations of the `chainaccess` interfaces as well.

The `chainlink-ccv/platform` module can then have the subdirectories `devenv` for the devenv package and `deployment` for the offchain deployment API?

deployment/adapters/aggregator_config.go
```diff
@@ -0,0 +1,89 @@
...
+	// ScanCommitteeStates returns committee states for the given chain from the deployment env.
+	ScanCommitteeStates(ctx context.Context, env deployment.Environment, chainSelector uint64) ([]*CommitteeState, error)
+	// ResolveVerifierAddress returns the verifier contract address for the given chain and qualifier using the datastore.
+	ResolveVerifierAddress(ds datastore.DataStore, chainSelector uint64, qualifier string) (string, error)
+}
+
+type AggregatorConfigRegistry struct {
+	mu       sync.Mutex
+	adapters map[string]AggregatorConfigAdapter
+}
```
makramkd: A developer ergonomics question: have we thought about potentially combining the registries into one, that ends up registering a `struct` of all these different interface types, in order to have a single entrypoint?

```golang
type Adapters struct {
    AggregatorConfigAdapter AggregatorConfigAdapter
    ExecutorConfigAdapter ExecutorConfigAdapter
    // the rest of them ... 
}

type AdapterRegistry struct {
    mu sync.Mutex
    adapters map[string]Adapters
}

func (r *AdapterRegistry) Register(family string, adapters Adapters) {
    // register func same as below.
}
```

deployment/topology.go
```diff
@@ -0,0 +1,381 @@
...
+)
+
+// EnvironmentTopology holds all environment-specific configuration that cannot be inferred
+// from the datastore. This serves as the single source of truth for the desired state of both off-chain
+// (job specs) and on-chain (committee contracts) configuration.
+type EnvironmentTopology struct {
+	IndexerAddress []string                      `toml:"indexer_address"`
+	PyroscopeURL   string                        `toml:"pyroscope_url"`
+	Monitoring     MonitoringConfig              `toml:"monitoring"`
+	NOPTopology    *NOPTopology                  `toml:"nop_topology"`
```
makramkd, line 24: This being a pointer suggests it could be `nil` (and we'd have to do `nil` checks in a lot of places). Is this the case or is it actually required?

deployment/topology.go
```diff
@@ -0,0 +1,381 @@
...
+	ExecutorPools  map[string]ExecutorPoolConfig `toml:"executor_pools"`
+}
+
+// NOPTopology defines the node operator structure and committee membership.
+type NOPTopology struct {
+	NOPs       []NOPConfig                `toml:"nops"`
+	Committees map[string]CommitteeConfig `toml:"committees"`
```
makramkd, line 31: This `NOPTopology` type seems weird to me because its committee-verifier specific is it not? Or is the entire `EnvironmentTopology` structure committee-verifier specific?

deployment/topology.go
```diff
@@ -0,0 +1,381 @@
...
+// ExecutorPoolConfig defines executor pool membership and configuration.
+type ExecutorPoolConfig struct {
+	ChainConfigs      map[string]ChainExecutorPoolConfig `toml:"chain_configs"`
+	IndexerQueryLimit uint64                             `toml:"indexer_query_limit"`
+	BackoffDuration   time.Duration                      `toml:"backoff_duration"`
+	LookbackWindow    time.Duration                      `toml:"lookback_window"`
+	ReaderCacheExpiry time.Duration                      `toml:"reader_cache_expiry"`
+	MaxRetryDuration  time.Duration                      `toml:"max_retry_duration"`
+	WorkerCount       int                                `toml:"worker_count"`
+	NtpServer         string                             `toml:"ntp_server"`
```
makramkd, line 129: These are the previously duplicated executor configs? Or were those somewhere else?

deployment/shared/types.go
```diff
@@ -0,0 +1,215 @@
...
+type BeholderInput struct {
+	InsecureConnection       bool
+	CACertFile               string
+	OtelExporterGRPCEndpoint string
+	OtelExporterHTTPEndpoint string
+	LogStreamingEnabled      bool
+	MetricReaderInterval     int64
+	TraceSampleRatio         float64
+	TraceBatchTimeout        int64
+}
```
makramkd: This seems to be duplicated in multiple places, can we import it from somewhere where its canonically defined?

deployment/aggregator.go
```diff
@@ -0,0 +1,20 @@
...
+type QuorumConfig struct {
+	SourceVerifierAddress string   `json:"sourceVerifierAddress"`
+	Signers               []Signer `json:"signers"`
+	Threshold             uint8    `json:"threshold"`
+}
+
+type Committee struct {
+	QuorumConfigs        map[SourceSelector]*QuorumConfig      `json:"quorumConfigs"`
+	DestinationVerifiers map[DestinationSelector]string `json:"destinationVerifiers"`
+}
```
makramkd, line 20: Duplicated configs?

deployment/indexer.go
```diff
@@ -0,0 +1,10 @@
+package deployment
+
+type IndexerVerifierConfig struct {
+	Name            string   `json:"name"`
+	IssuerAddresses []string `json:"issuerAddresses"`
+}
+
+type IndexerGeneratedConfig struct {
+	Verifiers []IndexerVerifierConfig `json:"verifiers"`
+}
```
makramkd, line 10: Ditto here, duplicated? Whats the plan in a follow up?

deployment/token_verifier.go
```diff
@@ -0,0 +1,55 @@
...
+
+import "time"
+
+type TokenVerifierGeneratedConfig struct {
+	PyroscopeURL       string                        `json:"pyroscope_url"`
+	OnRampAddresses    map[string]string             `json:"on_ramp_addresses"`
+	RMNRemoteAddresses map[string]string             `json:"rmn_remote_addresses"`
+	TokenVerifiers     []TokenVerifierEntry          `json:"token_verifiers"`
+	Monitoring         TokenVerifierMonitoringConfig `json:"monitoring"`
+}
```
makramkd, line 11: Ditto here, duplicated

makramkd (commented):
Pros: chain-agnostic interfaces, following existing patterns. Cons: config struct duplication, EnvironmentTopology seems committee-verifier specific, I feel like it either needs a better home or there needs to be some synergy b/w lane definitions, chain support, and committee verifier configurations, as all have been referred to as "topology" at some point in time and confusion is nigh upon us.

deployment/adapters/aggregator_config.go
tt-cll: Yeah I remember us talking about this. We can switch these to a single register to start the pattern. This was mostly a lift and shift from what already existed

deployment/indexer.go
tt-cll: We need to clean this up in general.

deployment/adapters/aggregator_config.go
makramkd: 🏋️

makramkd (approved):

carte7000 (approved):

PR #1038, opened 2026-04-21
Family Agnostic Standalone Executor

winder:
This PR supports a family agnostic executor by:
1. Adding executor services to the Accessor.
2. Updating GenericConfig with configs needed by the new services.
3. Updating the EVM AccessorFactory to build the new services.
4. Using the registry in cmd/executor/standalone/main.go instead of manually creating EVM objects.


cmd/executor/standalone/main.go
```diff
@@ -35,7 +32,6 @@ import (
 
 const (
 	configPathEnvVar = "EXECUTOR_CONFIG_PATH"
-	privateKeyEnvVar = "EXECUTOR_TRANSMITTER_PRIVATE_KEY"
```
winder, line 38: There should be a better way to inject this into the executor.

cmd/executor/standalone/main.go
```diff
@@ -346,15 +317,25 @@ func main() {
...
 	normalizedConfig, err := config.GetNormalizedConfig()
 	if err != nil {
 		return nil, nil, err
 	}
-	return normalizedConfig, &config.BlockchainInfos, nil
+
+	registry, err := chainaccess.NewRegistry(lggr, string(content))
+	if err != nil {
+		return nil, nil, err
+	}
```
winder, line 339: The main change. Create the registry, then use it in place of the other constructor calls that used to be in this file.

pkg/chainaccess/registry.go
```diff
@@ -69,6 +70,7 @@ type GenericConfig struct {
 	ChainConfig Infos[any] `toml:"blockchain_infos"`
 
 	CommitteeConfig
+	ExecutorConfig
```
winder, line 73: New overlay config. This defines the offramp/rmn addresses and other config used by the new services.

cmd/executor/standalone/main.go
makramkd: In the bootstrap approach this would be provided by the keystore

makramkd (approved):

tt-cll (approved):

PR #1040, opened 2026-04-22
switch messageOptions to an interface

0xAustinWang:
The previous version of our "MessageOptions" was an inflexible concrete struct that was difficult to extend for nonevms. 

This PR replaces the concrete struct with a `ExtraArgsDataProvider` interface and creates an example for SVM. It adds the serialization to be source chain specific. 

If we want to extend this in the future, we can add a hook to the extra arg data provider interface so implementers can add some logic.


build/devenv/evm/extraArgs.go
```diff
@@ -0,0 +1,40 @@
...
+	ExecutorArgs []byte
+	// TokenArgs are the token arguments for the message
+	TokenArgs []byte
+	// UseTestRouter when true looks up the TestRouter contract type in the datastore instead of Router.
+	UseTestRouter bool
+}
+
+func (m MessageOptions) IsExtraArgsDataProvider() {}
+
+// TODO: import these from Solana family.
```
0xAustinWang: This is a bit weird to import; it comes in as `evm.SVMMessageOptions`, maybe we can move these types into a registry (if it's not too annoying to access the registry from tests)

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -417,21 +404,25 @@ type ChainAsDestination interface {
 // Chain families can implement this interface to run partial CCIP message tests without having to implement the full `Chain` interface.
 type ChainAsSource interface {
 	genericChain
+	// ExtraArgsSerializer serializes the extra args for the given destination chain.
+	// Implementation should type assert the ExtraArgsDataProvider to struct types from supported destination chain families.
+	ExtraArgsSerializer(ExtraArgsDataProvider) ([]byte, error)
```
0xAustinWang: There's a bit of redundancy in the registry pattern when it lives alongside this interface. 

We register a serializer based on chain selector, but we can also use the serializer directly from the chainAsSource implementation

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -405,6 +389,9 @@ type genericChain interface {
 // Chain families can implement this interface to run partial CCIP message tests without having to implement the full `Chain` interface.
 type ChainAsDestination interface {
 	genericChain
+	// ExtraArgsProvider returns the extra-args data provider for this destination chain.
+	// The output of this method will be passed to the ExtraArgsEncoder in ChainAsSource.
+	// ExtraArgsProvider(any) (ExtraArgsDataProvider, error)
```
makramkd: The intention is that this gets added right? O/w we won't be able to have chain-agnostic tests (in evmPOC_test.go we currently pass in the evm.MessageOptions struct)

build/devenv/cciptestinterfaces/interface.go
makramkd: Naming nit: the -er at the end suggests an object being returned that does the serializing, but in this case the ChainAsSource serializes things, so maybe a more appropriate name is `SerializeExtraArgs`?

build/devenv/evm/extraArgs.go
makramkd: I actually don't think this should be in the `evm` package; the reason is that the SVM extra args are the same irrespective of whether you send from EVM or any other chain family. So this should be thought of as a generic Go struct IMO.

build/devenv/evm/extraArgs.go
```diff
@@ -0,0 +1,40 @@
+package evm
+
+import "github.com/smartcontractkit/chainlink-ccv/protocol"
+
+// MessageOptions consists of all the ways one can modify a CCIP message
+// using extraArgs.
+type MessageOptions struct {
```
makramkd: Hmm I think this `MessageOptions` starts to become confusing in the current setup because it doesn't correspond to a single ExtraArgs version but potentially all 3 versions (V1 legacy, generic V2, generic V3).

Maybe we should deprecate this and instead follow onchain more closely so that there isn't confusion also in the test suite.

This also makes me think, to support chain-agnostic test that test different extra args versions, the `ExtraArgsProvider` should probably take in, similar to the 1.6 implementation, a varargs of `type ExtraArgsOpt struct { Name string; Value any; }` where one of these options is the extra args version.

Then there's got to be some kind of lookup that is potentially chain-family specific - i.e. if i pass version == 1 to the SVM ChainAsDestination, it should return SVMMessageOptions (IMO should be renamed SVMExtraArgsV1), but if pass version == 2 it should fail (not supported on SVM) and if i pass version == 3 it should return GenericExtraArgsV3.

makramkd (commented):
I believe that the pattern being employed is fine but there's some details around how to organize the ExtraArgs. 

In my mind it should be its own package that is chain agnostic - I think we can put it in `cciptestinterfaces/extraargs` for now unless we find a better home. 

The reasoning is that if a destination chain can decide what to return as an ExtraArgsProvider, it would be returning that same struct for all potential source chain families. 

For a concrete example, if the destination is SVM, the SVM ChainAsDestination implementation would return a concrete struct SVMExtraArgsV1 implementing the marker interface. All potential source chains need to handle it with a type switch `switch provider.(type)` and potentially serialize it differently (i.e. on EVM it uses different extra args selectors) if they support SVM as a destination chain family in the onchain code.

build/devenv/evm/extraArgs.go
0xAustinWang: Will add extra args versioning in a future PR. This one is scoped to reducing reliance on the concrete struct. Once we move towards an interface, implementers can do anything they want with it

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+}
+
+func WithTokenArgs(args []byte) ExtraArgsOption {
+	return func(p ExtraArgsDataProvider) error {
+		return mutateMessageOptions("evm.WithTokenArgs", p, func(m *MessageOptions) { m.TokenArgs = args })
+	}
+}
+
+func WithUseTestRouter(b bool) ExtraArgsOption {
```
winder: Nit: the boolean options could be shortened to `WithTestRouter()` and `WithOrderedExecution()`. Then you simply omit them for the standard router and unordered execution.

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,116 @@
...
+
+// ExtraArgsOption mutates a chain-specific ExtraArgsDataProvider that has been
+// allocated by a destination chain's ExtraArgsBuilder. Each chain family defines
+// its own option constructors (e.g. evm.WithExecutionGasLimit) that type-assert
+// the provider to the concrete struct; applying an option to the wrong chain
+// family returns an error rather than silently no-op'ing.
+type ExtraArgsOption func(ExtraArgsDataProvider) error
+
+// MessageOptions represents EVM modifications one can make to a CCIP message for through extra args.
+type MessageOptions struct {
```
huangzhen1997, line 9: In our current evm2solana e2e test we used [SVMExtraArgsV1](https://github.com/smartcontractkit/chainlink-ccv/blob/fb3294b643fe548d0c3ca39fc8f3ce75bfd6649a/build/devenv/evm/impl.go#L699) from 1.6, it worked because our offramp is still in development, so the offramp is not validating the msg from source chain. But in the future when constructing ccip send msg, we would reuse this MessageOptions here (as [GenericExtraArgsV3](https://github.com/smartcontractkit/chainlink-ccip/blob/0d0f33d260382eec1b485f872b2e7a2af1e17eb5/chains/evm/contracts/libraries/ExtraArgsCodec.sol#L73)), and use the Solana specific [executorArgs](https://github.com/smartcontractkit/chainlink-ccip/blob/0d0f33d260382eec1b485f872b2e7a2af1e17eb5/chains/evm/contracts/libraries/ExtraArgsCodec.sol#L157). cc @toblich @archseer

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/smartcontractkit/chainlink-ccv/protocol"
+)
+
+// ExtraArgsOption mutates a chain-specific ExtraArgsDataProvider that has been
+// allocated by a destination chain's ExtraArgsBuilder. Each chain family defines
+// its own option constructors (e.g. evm.WithExecutionGasLimit) that type-assert
+// the provider to the concrete struct; applying an option to the wrong chain
+// family returns an error rather than silently no-op'ing.
+type ExtraArgsOption func(ExtraArgsDataProvider) error
```
makramkd: Is the option approach here the appropriate one vs. what we have in 1.6 (https://github.com/smartcontractkit/chainlink-ccip/blob/1925e2885bbec526b9e05c01260b74efe4ec56b4/deployment/testadapters/adapters.go#L56-L62)

The way I'm imagining we'll be writing tests that change extra args is sort of similar to today but w/out specifying MessageOptions{} explicitly because that doesn't cover all extra args - only V3. We also have chain-family specific fields in the V3 extra args, specifically the executor extra args (e.g. solana https://github.com/smartcontractkit/chainlink-ccip/blob/1925e2885bbec526b9e05c01260b74efe4ec56b4/chains/evm/contracts/libraries/ExtraArgsCodec.sol#L157-L163 and sui https://github.com/smartcontractkit/chainlink-ccip/blob/1925e2885bbec526b9e05c01260b74efe4ec56b4/chains/evm/contracts/libraries/ExtraArgsCodec.sol#L165-L167).

So, there are mostly generic fields and a chain-family specific field (`executorArgs`), at least for the V3 extra args.

```golang
type ExtraArgOpt struct {
	Name  string
	Value any
}

func NewV3ExtraArg() ExtraArgOpt {
  return ExtraArgOpt{
    Name: "version", // can pull out to a constant so BuildExtraArgs implementations can check for it
    Value: 3,
  }
}

func NewCCVsExtraArg(ccvs []protocol.UnknownAddress) ExtraArgOpt {
  return ExtraArgOpt{
    Name: "ccvs", // ditto here re: constant
    Value: ccvs,
  }
}

type AbstractExtraArgs any

// BuildExtraArgsV2OrV3 handles all the generic options in the V2 and V3 extra args,
// and leaves executorArgs empty, to be handled by chain-family implementations.
func BuildExtraArgsV2OrV3(opts ...ExtraArgOpt) (AbstractExtraArgs, error) {
  // this function processes opts and returns either
  // GenericExtraArgsV2 or GenericExtraArgsV3 depending on the version
  // and the remaining options.
  // if neither of these versions are specified, it returns ErrUnsupportedVersion
}

// on the implementations of BuildExtraArgs on ChainAsDestination,
// the following logic exists
func (a *CASImpl) BuildExtraArgs(opts ...cciptestinterfaces.ExtraArgOpt) (AbstractExtraArgs, error) {
  // if v2 or v3, use helper
  extraArgs, err := cciptestinterfaces.BuildExtraArgsV2OrV3(opts)
  if err == nil {
    // for solana, we need to update executor args appropriately
    // see https://github.com/smartcontractkit/chainlink-ccip/blob/1925e2885bbec526b9e05c01260b74efe4ec56b4/chains/evm/contracts/libraries/ExtraArgsCodec.sol#L157-L163
    // set the field appropriately then return
    return extraArgs, nil
  }
  // err is non-nil
  if err != ErrUnsupportedVersion {
    // some other error, return early
    return nil, fmt.Errorf("error processing extra args options: %w", err)
  }

  // else, process manually, e.g. SVMExtraArgsV1
  versionOpt, err := getVersionOpt(opts)
  if versionOpt.Value != 1 {
    return nil, fmt.Errorf("expected version 1 for SVM extra args")
  }

  // build SVMExtraArgsV1 and return based on opts, e.g computeUnits etc. and return
}

src, ok := srcImpl.(ChainAsSource)
dst, ok := dstImpl.(ChainAsDestination)

// This test suite always uses v3, for example
for _, ccvsCombo := range allCCVCombos {
  extraArgs, err := dst.BuildExtraArgs(NewV3ExtraArg(), NewCCVsExtraArg(ccvCombo))
  require.NoError(t, err)
}

// Solana-specific suite, uses V1
extraArgs, err := dst.BuildExtraArgs(ExtraArgsOpt{Name: "version", Value: 1}, ExtraArgsOpt{Name: "computeUnits", Value: 50_000}, ...)
require.NoError(t, err)
```

0xAustinWang: I tried this approach out but ultimately decided to move towards a more idiomatic approach using explicit interfaces per version rather than building something through abstract types. 

Specifically, chain family and message version are always defined and used to determine which "serializer" to utilize. We don't need a high level generic "BuildExtraArgs"; we can use lower level serializeEVMv2, serializeSVMv1, etc. 

This approach works well with the milestone based approach because we determine which messages have been implemented based on type casting into different interfaces.

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,103 @@
...
+	BuildV3ExtraArgs(
+		opts MessageOptions,
+		destChain MessageV3Destination,
+		executorArgsParams any,
+		tokenArgsParams any,
+	) ([]byte, error)
+}
+
+// Any2EVMMessageV2Data represents the data for V2 messages arriving at an EVM chain.
+type Any2EVMMessageV2Data struct {
```
makramkd: I don't think this should have `EVM` in the name, this is `GenericExtraArgsV2` which applies to many chain families?

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,103 @@
...
+
+// ExtraArgsOption mutates a chain-specific ExtraArgsDataProvider that has been
+// allocated by a destination chain's ExtraArgsBuilder. Each chain family defines
+// its own option constructors (e.g. evm.WithExecutionGasLimit) that type-assert
+// the provider to the concrete struct; applying an option to the wrong chain
+// family returns an error rather than silently no-op'ing.
+type ExtraArgsOption func(ExtraArgsDataProvider) error
+
+// MessageOptions represents modifications one can make to a CCIP messagev3 through extra args.
+// TODO: rename this to Any2AnyMessageV3Data.
```
makramkd: We've ditched the "Any2X" or "X2Any" nomenclature for V2, I don't think we should revisit it else would cause confusion - `GenericExtraArgsV3` seems better to me.

makramkd (dismissed):

carte7000 (dismissed):

build/devenv/cciptestinterfaces/extra_args.go
0xAustinWang: I think we still need some sort of chain distinction. Messages arriving on EVM have this data
```
{
	GasLimit                 uint32
	AllowOutOfOrderExecution bool
}
```
while messages arriving on SVM have this data
```
{
	Version                  uint8
	ComputeUnits             uint32
	AccountIsWritableBitmap  uint64
	AllowOutOfOrderExecution bool
	TokenReceiver            [32]byte
	Accounts                 [][32]byte
}
```

They're both classified as "V2" messages, but the contents are different

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/smartcontractkit/chainlink-ccv/protocol"
+)
+
+// MessageOptions represents modifications one can make to a CCIP messagev3 through extra args.
+// TODO: rename this to GenericExtraArgsV3.
+type MessageOptions struct {
+	// ExecutionGasLimit is the execution gas limit for the message
+	ExecutionGasLimit uint32
+	// OutOfOrderExecution is whether to execute the message out of order
+	OutOfOrderExecution bool
```
makramkd, line 14: **suggestion**: it seems like `OutOfOrderExecution` is not part of GenericExtraArgsV3, should it be removed from this struct pre-emptively? Or put a TODO next to it for when we do rename?

build/devenv/cciptestinterfaces/extra_args.go
makramkd: This is a nitpick but the SVM you outlined is named "SVMExtraArgsV1" in the onchain code.

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,96 @@
...
+	GasLimit                 uint32
+	AllowOutOfOrderExecution bool
+}
+
+// This fulfills the marker interface ExtraArgsDataProvider.
+func (m Any2EVMMessageV2Data) IsExtraArgsDataProvider() {}
+
+// Any2EVMMessageV2 is an interface for any chain that can send a V2 message to an EVM chain.
+type Any2EVMMessageV2 interface {
+	SerializeAny2EVMMessageV2(opts any) ([]byte, error)
```
makramkd: **nitpick**: The V3 method is called `BuildV3ExtraArgs`, should we follow a similar naming pattern here, i.e. `BuildGenericV2ExtraArgs`?

build/devenv/cciptestinterfaces/extra_args.go
makramkd: Its also defined below as `Any2SVMMessageV1Data` - so the distinction seems to already be there.

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Any2EVMMessageV1Data represents the data for V1 messages arriving at an EVM chain.
+type Any2EVMMessageV1Data struct {
```
makramkd: **suggestion**: name it `EVMExtraArgsV1` to mirror onchain struct?

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,96 @@
...
+	GasLimit uint32
+}
+
+// This fulfills the marker interface ExtraArgsDataProvider.
+func (m Any2EVMMessageV1Data) IsExtraArgsDataProvider() {}
+
+// Any2EVMMessageV1 is an interface for any chain that can send a V1 message to an EVM chain.
+type Any2EVMMessageV1 interface {
+	SerializeAny2EVMMessageV1(opts any) ([]byte, error)
```
makramkd: **nitpick**: similar to above, name could be `BuildEVMExtraArgsV1`?

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Any2SVMMessageV1Data represents the data for V1 messages arriving at a SVM chain.
+type Any2SVMMessageV1Data struct {
```
makramkd: **suggestion**: name it `SVMExtraArgsV1` to match onchain struct?

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,96 @@
...
+	TokenReceiver            [32]byte
+	Accounts                 [][32]byte
+}
+
+// This fulfills the marker interface ExtraArgsDataProvider.
+func (m Any2SVMMessageV1Data) IsExtraArgsDataProvider() {}
+
+// Any2SVMMessageV1 is an interface for any chain that can send a V1 message to a SVM chain.
+type Any2SVMMessageV1 interface {
+	SerializeAny2SVMMessageV1(opts any) ([]byte, error)
```
makramkd: **nitpick**: similar to above, name it `BuildSVMExtraArgsV1`? This current naming makes it sound like you're serializing the whole message, which it isn't, right?

build/devenv/tests/composable/messaging/agnostic_chain_test.go
```diff
@@ -0,0 +1,106 @@
...
+	dest cciptestinterfaces.ChainAsDestination,
+	fields cciptestinterfaces.MessageFields,
+	opts cciptestinterfaces.Any2EVMMessageV2Data,
+	sendOption cciptestinterfaces.ChainSendOption,
+) error {
+	v2Source, ok := source.(cciptestinterfaces.Any2EVMMessageV2)
+	if !ok {
+		return fmt.Errorf("source chain does not implement V2 message")
+	}
+	sourceExtraArgs, err := v2Source.SerializeAny2EVMMessageV2(opts)
```
makramkd: **question**: this test isn't chain agnostic, right? Whats the plan to make it chain agnostic?

carte7000 (dismissed):

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -0,0 +1,96 @@
...
+// We use an interface rather than a struct because the V3 message structure is chain agnostic.
+type MessageV3Source interface {
+	// BuildV3ExtraArgs builds the V3 extra arguments for the message including calling the destination chain's GetExecutorArgs and GetTokenArgs.
+	// then serializing the results into the source chain specific encoding format, and returning the result.
+	BuildV3ExtraArgs(
+		opts MessageOptions,
+		destChain MessageV3Destination,
+		executorArgsParams any,
+		tokenArgsParams any,
+	) ([]byte, error)
```
winder: nit: similar to `GenericChainMessage`, use a type to link this output to `BuildChainMessage
```suggestion
	) (GenericExtraArgs, error)
```

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -419,19 +409,20 @@ type ChainAsSource interface {
 	genericChain
```
winder, line 410: nit: can you get away with leaving the ChainSelector() helper out of the interface? Maybe it can be moved to `SendOptions` for implementations that need it?

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -419,19 +409,20 @@ type ChainAsSource interface {
...
 	// BuildChainMessage builds a CCIP message for the given destination chain.
 	// It will call into the registered extra args serializer per destination chain for now, until we have a more generic way to manage extra args.
-	// It returns a generic type that is specific to the chain family. The returned message is expected to be directly passed in ot the SendChainMessage method.
+	// It returns a generic type that is specific to the chain family. The returned message is expected to be directly passed in to the SendChainMessage method.
 	// For example, the EVM implementation returns a routerwrapper.ClientEVM2AnyMessage.
-	BuildChainMessage(ctx context.Context, destChain uint64, messageFields MessageFields, opts MessageOptions) (ChainAsSourceMessage, error)
+	BuildChainMessage(ctx context.Context, destChain uint64, messageFields MessageFields, extraArgs []byte) (GenericChainMessage, error)
```
winder: nit: looks like this isn't used for the EVM implementation. If possible to omit, I think it would help discourage per family switch statements if it isn't even part of the interface
```suggestion
	BuildChainMessage(ctx context.Context, messageFields MessageFields, extraArgs []byte) (GenericChainMessage, error)
```

build/devenv/tests/composable/messaging/agnostic_chain_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	extraArgs, err := v3Serializer.BuildV3ExtraArgs(opts, v3Receiver, executorArgsParams, tokenArgsParams)
+	if err != nil {
+		return fmt.Errorf("failed to encode V3 extra args: %w", err)
+	}
+
+	srcMessage, err := srcChain.BuildChainMessage(ctx, destChain.ChainSelector(), fields, extraArgs)
+	if err != nil {
+		return fmt.Errorf("failed to build chain message: %w", err)
+	}
```
winder: The `v3Serializer` and `srcChain` are the same object which makes me wonder if these can be combined into one function.

```
BuildChainMessage(context.Context, MessageFields, ExtraArgsProvider, ExecutorArgs, TokenArgs)
```

build/devenv/tests/composable/messaging/agnostic_chain_test.go
```diff
@@ -0,0 +1,106 @@
...
+	v3Receiver, ok := destChain.(cciptestinterfaces.MessageV3Destination)
+	if !ok {
+		return fmt.Errorf("dest chain does not support V3 message")
+	}
+	v3Serializer, ok := srcChain.(cciptestinterfaces.MessageV3Source)
+	if !ok {
+		return fmt.Errorf("source chain does not support V3 message")
+	}
+
+	extraArgs, err := v3Serializer.BuildV3ExtraArgs(opts, v3Receiver, executorArgsParams, tokenArgsParams)
```
winder: I thought extraArgs and tokenArgs were regular message fields

build/devenv/evm/impl.go
```diff
@@ -117,13 +117,19 @@ func init() {
...
 	// Canton shares EVM's extra args serialization. Canton's product repo can
 	// register its own serializer if the formats ever diverge; until then, this
 	// provides backward compatibility with the previous FamilyEVM/FamilyCanton
 	// combined switch case.
-	cciptestinterfaces.RegisterExtraArgsSerializer(chainsel.FamilyCanton, SerializeEVMExtraArgs)
-	cciptestinterfaces.RegisterExtraArgsSerializer(chainsel.FamilySolana, SerializeSVMExtraArgs)
+
+	cciptestinterfaces.RegisterExtraArgsSerializer(cciptestinterfaces.ExtraArgsSerializerEntry{Family: chainsel.FamilyCanton, Version: 3}, SerializeMessageV3ExtraArgs)
+	cciptestinterfaces.RegisterExtraArgsSerializer(cciptestinterfaces.ExtraArgsSerializerEntry{Family: chainsel.FamilyCanton, Version: 2}, SerializeAny2EVMMessageV2)
+	cciptestinterfaces.RegisterExtraArgsSerializer(cciptestinterfaces.ExtraArgsSerializerEntry{Family: chainsel.FamilyCanton, Version: 1}, SerializeAny2EVMMessageV1)
```
winder: Messages to canton are serialized the same way as messages to EVM?

build/devenv/tests/composable/messaging/agnostic_chain_test.go
```diff
@@ -0,0 +1,106 @@
...
+	sendOption cciptestinterfaces.ChainSendOption,
+	executorArgsParams any,
+	tokenArgsParams any,
+) error {
+	v3Receiver, ok := destChain.(cciptestinterfaces.MessageV3Destination)
+	if !ok {
+		return fmt.Errorf("dest chain does not support V3 message")
+	}
+	v3Serializer, ok := srcChain.(cciptestinterfaces.MessageV3Source)
+	if !ok {
```
winder, line 30: Consistent names would be helpful
```suggestion
	v3Dest, ok := destChain.(cciptestinterfaces.MessageV3Destination)
	if !ok {
		return fmt.Errorf("dest chain does not support V3 message")
	}
	v3Source, ok := srcChain.(cciptestinterfaces.MessageV3Source)
```

winder (dismissed):
I checked out the code and find the implementations of the interface and the implementations using the interface both look really nice.

The actual interface itself is still a bit hard for me to wrap my head around. I wonder about how people would discover that they can cast `ChainAsSource` to  `MessageV3Source` in order to start using finality fields.

Overall I'm fine with that because this PR looks like a great improvement towards making message assembly more modular. If we find ways to make specific usage patterns more convenient they can be simple helpers ontop of the low level primitives.

build/devenv/tests/composable/messaging/agnostic_chain_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/smartcontractkit/chainlink-ccv/build/devenv/cciptestinterfaces"
+)
+
+// MessageV3TestScenario is a helper test scenario that sends a CCIP message using the ChainAsSource and ChainAsDestination interfaces.
+// It uses these partial interfaces to build, send, and assert messages in a way that is reusable across different chain families.
+// After a chain integration implements either ChainAsSource or ChainAsDestination, they can wire up this test scenario to verify their implementation (evmPOC_test.go is an example).
+// This test scenario supports partial implementation, so an integration can implement source side FIRST, connecting it to the existing EVM destination to confirm source side functionality.
+func MessageV3TestScenario(
+	ctx context.Context,
+	t *testing.T,
```
winder: nit: `testing.T` is unused.

build/devenv/evm/impl.go
0xAustinWang: I just kept the same behavior as what we had previously

build/devenv/tests/composable/messaging/agnostic_chain_test.go
0xAustinWang: There are message level token args that determine amount and token address, but the v3 extra args also passes arguments through

https://github.com/smartcontractkit/chainlink-ccip/blob/ce413dbadde1fbf0849cadb38d953eb1bb35a8a5/chains/evm/contracts/libraries/ExtraArgsCodec.sol#L115

build/devenv/tests/composable/messaging/agnostic_chain_test.go
0xAustinWang: Calling it chain agnostic makes sense because the consumer is a generic altvm team who only cares about sending messages AS SOURCE. Not all integrations will care about sending messages to aptos for example. 

For now, I'm adding a separate test to support SVM as a destination. 

I had previously investigated using a generic type here to indicate what shape the extra args should be, but this PR is getting too large

build/devenv/tests/composable/messaging/agnostic_chain_test.go
0xAustinWang: I thought about this, but didn't want to pass executorArgs/tokenArgs to the function if they're not supporting v3. The other approach is to have interfaces per chain family/message type ie 
SVMV1MessageBuilder/AptosV1MessageBuilder/GenericV3MessageBuilder

I don't think it's necessary now, but if the boilerplate gets too clunky in the future we can migrate to it.

winder (approved):

makramkd (approved):

PR #1041, opened 2026-04-22
deployment lint fixes

winder:
Run `lint fix` to pick up some problems in the deployment package. It seems that the lint job wasn't properly returning errors, so these slipped in.

PR #1042, opened 2026-04-22
re-enable deployment lint

tt-cll:
follow-up of https://github.com/smartcontractkit/chainlink-ccv/pull/1041

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -49,7 +49,7 @@ fmt: ensure-golangci-lint
 
 # Run golangci-lint
 lint fix="": ensure-golangci-lint
-    #gomods -c 'golangci-lint run {{ if fix != "" { "--fix" } else { "" } }}'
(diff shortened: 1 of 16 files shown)


makramkd (dismissed):

winder (dismissed):

Justfile
```diff
@@ -49,13 +49,12 @@ fmt: ensure-golangci-lint
 
 # Run golangci-lint
 lint fix="": ensure-golangci-lint
-    #gomods -c 'golangci-lint run {{ if fix != "" { "--fix" } else { "" } }}'
+    gomods -c 'golangci-lint run {{ if fix != "" { "--fix" } else { "" } }}'
```
winder: Technically this runs `golangci-lint` in every go.mod directory, so it's not necessary. But it also uses the wrong config. Maybe I can sneak in `--config` with an absolute path and turn this back into a 1 liner.

makramkd (approved):

winder (approved):

PR #1044, opened 2026-04-23
Migrate executor standalone to bootstrap.Run

winder:
Migrates the executor standalone binary to use bootstrap.Run, the same lifecycle manager used by all other CCV services. The old main.go was a ~230-line flat function with inline signal handling, OTEL setup, and all coordinator wiring. It is now minimal boilerplate (~30 lines); all executor-specific startup and shutdown logic lives in cmd/executor/service.go.

Two bugs fixed along the way:
  - bootstrap.Stop panicked on nil lifecycleManager/infoServer when using WithTOMLAppConfig (no JD). Nil guards added.
  - EVMContractTransmitter stored the startup context (10 s TTL from bootstrap.Run) and reused it for every PendingNonceAt/SuggestGasPrice call, causing silent context deadline exceeded failures after 10s. Context is now passed per-call through ConvertAndWriteMessageToChain.

See changelog/2026-04-24_executor_bootstrap.md for migration instructions.


integration/pkg/contracttransmitter/evm_contract_transmitter.go
```diff
@@ -112,7 +110,7 @@ func (ct *EVMContractTransmitter) ConvertAndWriteMessageToChain(ctx context.Cont
 	for _, ccv := range report.CCVS {
 		contractCcvs = append(contractCcvs, common.HexToAddress(ccv.String()))
 	}
-	opts, err := ct.GetTransactOpts()
+	opts, err := ct.GetTransactOpts(ctx)
```
winder, line 113: There's a sneaky bug here. Previously `GetTransactOpts()` used `context.Background()`, passed in through `NewEVMContractTransmitterFromRPC`. When that function is called by bootstrap, it uses a real context scoped to `GetAccessor` (given a generous 10 second duration context).

This leads to a situation where after 10 seconds, all contract transmits are cancelled.

winder: note: this bug doesn't impact the CL node deployment because the TXM based transmitter doesn't use a context in the same way.

0xAustinWang (dismissed):

bootstrap/bootstrap.go
```diff
@@ -233,13 +233,22 @@ func (b *Bootstrapper) Start(ctx context.Context) error {
...
-		return fmt.Errorf("failed to stop lifecycle manager: %w", err)
+	if b.lifecycleManager != nil {
+		if err := b.lifecycleManager.Stop(); err != nil {
+			return fmt.Errorf("failed to stop lifecycle manager: %w", err)
+		}
+	}
+	if b.infoServer != nil {
+		if err := b.infoServer.Stop(ctx); err != nil {
+			return fmt.Errorf("failed to stop info server: %w", err)
+		}
```
winder, line 251: Silly bug introduced with the static config option. These objects might be nil, so Stop() would always panic.

makramkd (dismissed):

integration/pkg/contracttransmitter/evm_contract_transmitter.go
```diff
@@ -77,20 +75,20 @@ func NewEVMContractTransmitterFromRPC(ctx context.Context, lggr logger.Logger, c
...
 	fromAddress := crypto.PubkeyToAddress(*publicKeyECDSA)
-	nonce, err := ct.Client.PendingNonceAt(ct.ctx, fromAddress)
+	nonce, err := ct.Client.PendingNonceAt(ctx, fromAddress)
 	if err != nil {
 		return nil, err
 	}
 
-	gasPrice, err := ct.Client.SuggestGasPrice(ct.ctx)
+	gasPrice, err := ct.Client.SuggestGasPrice(ctx)
 	if err != nil {
```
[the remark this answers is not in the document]
winder: Yes, I'm aware.

carte7000 (approved):

tt-cll (approved):

PR #1051, opened 2026-04-27
poc: std HeadTracker and RMNCurseReader conformance

makramkd:
This PR is a PoC of a suite of conformance tests for some accessor interfaces (currently just RMNCurseReader and HeadTracker).

This was me guiding the AI through a lot of it, but I think there can be improvements, mainly around e.g. the RMNCurseReader, since we might be able to use the standard `CurseAdapter` from the deployment API instead of the custom harness for this suite, but that would look a bit complex for the PoC.

I provided samples of how the conformance suites would be run for the existing EVM source reader.

### DIFF
--- a/pkg/chainaccess/rmncurseconformance/conformance.go
+++ b/pkg/chainaccess/rmncurseconformance/conformance.go
@@ -1,0 +1,125 @@
+package rmncurseconformance
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/smartcontractkit/chainlink-ccv/pkg/chainaccess"
+	"github.com/smartcontractkit/chainlink-ccv/protocol"
+)
+
+// RMNCurseHarness mutates on-chain (or in-chain) RMN Remote curse state for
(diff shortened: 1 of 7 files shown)


pkg/chainaccess/rmncurseconformance/conformance.go
```diff
@@ -0,0 +1,125 @@
...
+	// if a second deploy is not supported.
+	DeployRMN(ctx context.Context) (rmnAddress protocol.UnknownAddress, err error)
+	// CurseRMN curses the given subjects (e.g. chain selectors or global
+	// curse as [protocol.Bytes16]).
+	CurseRMN(ctx context.Context, subjects []protocol.Bytes16) error
+	// ClearRMNCurses removes all curses so a subsequent
+	// [chainaccess.RMNCurseReader.GetRMNCursedSubjects] returns the empty
+	// slice (or the chain’s pre-curse baseline). Safe to call when already clear.
+	ClearRMNCurses(ctx context.Context) error
+}
```
makramkd, line 30: Can probably be replaced by CurseAdapter but this is just a PoC - real implementation would use CurseAdapter to reduce duplication.

PR #1052, opened 2026-04-27
Migrate executor to JD-only lifecycle, remove standalone mode

winder:
The executor previously supported two launch paths: a "standalone" mode (static TOML config mounted into the container, no DB, no JD) and a "bootstrapped" mode (JD-managed, with its own Postgres DB). This PR removes standalone mode entirely — all executors now always use the bootstrap.Run / JD lifecycle.


build/devenv/env.toml
```diff
@@ -313,30 +313,36 @@ execution_interval = 15_000_000_000
 
 [[executor]]
   image = "executor:dev"
-  port = 8101
```
makramkd, line 316: Nice, yeah we need to clean up all these ad-hoc host port specifications, they shouldn't be managed manually by the user as its too error-prone.

changelog/2026-04-27_executor_jd_migration.md
```diff
@@ -0,0 +1,147 @@
...
+| `ImplFactory` interface | `SupportsBootstrapExecutor() bool` present | method removed |
+| `evmImplFactory` | returned `false` | method removed |
+| Devenv routing | two launch paths (`launchStandaloneExecutors` / `launchBootstrappedExecutors`) | single `launchExecutors` |
+
+**If you implement `ImplFactory` in a product repo**, remove the
+`SupportsBootstrapExecutor()` method.
+
+---
+
+## Breaking change: `executorsvc.NewStandalone` removed
```
makramkd, line 28: This actually made me think about the devenv API a bit, while technically this is a breaking change in the public API, this function should never be called by users. So we need some kind of `internal/`

changelog/2026-04-27_executor_jd_migration.md
```diff
@@ -0,0 +1,147 @@
...
+been removed. It duplicated executor container logic that now lives in
+`build/devenv/services/executor/base.go` (package `executorsvc`).
+
+If you imported types from `services.ExecutorInput`, `services.NewExecutor`, or
+`services.ApplyExecutorDefaults`, migrate to the equivalents in
+`build/devenv/services/executor/`.
+
+---
+
+## Breaking change: `env.toml` executor entries require a `[executor.db]` block
```
makramkd, line 69: This is a good callout - goes back to discussions about "versioning" of env.toml

changelog/2026-04-27_executor_jd_migration.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+After:
+```toml
+[[executor]]
+  container_name = "my-executor"
+  [executor.db]
+    image = "postgres:16-alpine"
+    name = "my-executor-db"
+    port = 8460
```
makramkd: Is the `port` field needed? I think we need a ticket to clean up all the fields that can be auto-assigned by docker (mostly ports and such)

makramkd (dismissed):

changelog/2026-04-27_executor_jd_migration.md
winder: Versioning for each top-level object would make sense, in this case the "[[executor]]" schema is changing.

makramkd (approved):

0xAustinWang (approved):

PR #1054, opened 2026-04-27
add threshold changesets + simplify evm offchain impl

tt-cll:


deployment/changesets/decrease_threshold.go
```diff
@@ -0,0 +1,188 @@
...
+	// ServiceIdentifier scopes the aggregator config DataStore output.
+	ServiceIdentifier string
+}
+
+// DecreaseThresholdOffchain is step-2 of the DecreaseThreshold two-entry product (§5.6).
+//
+// It asserts that the onchain threshold has already been lowered to ExpectedThreshold,
+// then regenerates the aggregator config to match. Triggered by the CLD post-proposal
+// hook after timelock execution.
+func DecreaseThresholdOffchain(registry *adapters.Registry) deployment.ChangeSetV2[DecreaseThresholdOffchainInput] {
```
carte7000, line 114: Worth noting that this will need to be run for every aggregator that exists in the environment.

carte7000: We could also consider specify all the serviceIdentifiers that need this config so that we update all of them at once?

deployment/changesets/increase_threshold.go
```diff
@@ -0,0 +1,188 @@
...
+			if err := outputDS.Merge(e.DataStore); err != nil {
+				return deployment.ChangesetOutput{}, fmt.Errorf("failed to merge datastore: %w", err)
+			}
+		}
+		if err := ccvdeployment.SaveAggregatorConfig(outputDS, cfg.ServiceIdentifier, committee); err != nil {
+			return deployment.ChangesetOutput{}, fmt.Errorf("failed to save aggregator config: %w", err)
+		}
+
+		// Step 2 (onchain): submit applySignatureConfigs with the updated threshold.
+		// In deployer-key mode this is a direct transaction submission.
```
carte7000: If the onchain goes first here this might cause issue? The aggregator config will need to be deployed first which is done in another repo. We likely need a 2 step approach like the decrease threshold here?

carte7000 (approved):

AnieeG (approved):

winder (approved):

PR #1057, opened 2026-04-28
executor: replace env-var private key with keystore-managed transmitter key

winder:
Replaces the raw private key (EXECUTOR_TRANSMITTER_PRIVATE_KEY / transmitter_private_key) with a keystore-managed ECDSA key declared at startup via bootstrap.WithKey. Key material never leaves the keystore.

Breaking changes:
  - evm.NewFactory: transmitterPrivateKey parameter removed
  - EXECUTOR_TRANSMITTER_PRIVATE_KEY env var no longer read
  - Executor key name: "default_evm_key" → "evm/tx/executor_evm_transmitter_key" (executor.DefaultEVMTransmitterKeyName)
  - ImplFactory: GenerateTransmitterKey and TransmitterAddress removed; related devenv helpers (SetTransmitterPrivateKey, TransmitterKeyGenerator, etc.) removed
  - chainaccess.Registry is now an interface; NewRegistry returns the interface
  - Key name constants moved: keys.DefaultCSAKeyName → bootstrap.DefaultCSAKeyName, keys.DefaultECDSASigningKeyName → commit.DefaultECDSASigningKeyName

New:
  - KeystoreEVMContractTransmitter — signs OffRamp transactions via keystore without exposing key material
  - bootstrap.KeystoreSetter — optional interface for Accessors that need a keystore; the EVM accessor implements it to lazily build the transmitter
  - bootstrap.KeystoreRegistry — wraps any Registry and automatically calls SetKeystore on accessors that support it; warns if an accessor doesn't implement the interface
  - DestinationChainConfig.TransmitterKeyName — optional per-chain TOML override for the transmitter key name

### DIFF
--- a/build/devenv/environment.go
+++ b/build/devenv/environment.go
@@ -2,6 +2,7 @@ package ccv
 
 import (
 	"context"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -1418,6 +1419,10 @@ func NewEnvironment() (in *Cfg, err error) {
 		return nil, fmt.Errorf("failed to create executors: %w", err)
 	}
 
+	if err := fundBootstrappedExecutors(ctx, in.Executor, in.Blockchains, impls); err != nil {
+		return nil, fmt.Errorf("failed to fund bootstrapped executor transmitters: %w", err)
+	}
+
 	if jdInfra != nil && jdInfra.OffchainClient != nil {
 		if err := registerExecutorsWithJD(ctx, in.Executor, jdInfra.OffchainClient); err != nil {
 			return nil, err
@@ -1783,6 +1788,59 @@ func launchCLNodes(
 	Plog.Info().Any("OnchainPublicKeys", onchainPublicKeys).Msg("Onchain public keys for all nodes")
 
 	return onchainPublicKeys, nil
+}
+
+// fundBootstrappedExecutors funds the EVM transmitter addresses of all bootstrapped
+// (non-standalone) executors that have an EVMTransmitterAddress in their BootstrapKeys.
+// Standalone executors are funded earlier via their TransmitterPrivateKey before launch.
+func fundBootstrappedExecutors(
+	ctx context.Context,
+	executors []*executorsvc.Input,
+	blockchains []*blockchain.Input,
+	impls []cciptestinterfaces.CCIP17Configuration,
+) error {
+	addressesByFamily := make(map[string][]protocol.UnknownAddress)
+	for _, exec := range executors {
+		if exec == nil || exec.Mode == services.Standalone {
+			continue
+		}
+		if exec.Out == nil || exec.Out.BootstrapKeys.EVMTransmitterAddress == "" {
+			continue
+		}
+		family := exec.ChainFamily
+		if family == "" {
+			family = chainsel.FamilyEVM
+		}
+		addrBytes, err := hex.DecodeString(exec.Out.BootstrapKeys.EVMTransmitterAddress)
+		if err != nil {
+			return fmt.Errorf("invalid EVM transmitter address for executor %s: %w", exec.ContainerName, err)
+		}
+		addressesByFamily[family] = append(addressesByFamily[family], protocol.UnknownAddress(addrBytes))
+	}
+
+	for i, impl := range impls {
+		if i >= len(blockchains) {
+			break
+		}
+		family, famErr := blockchain.TypeToFamily(blockchains[i].Type)
+		if famErr != nil {
+			continue
+		}
+		fac, facErr := GetImplFactory(string(family))
+		if facErr != nil || !fac.SupportsFunding() {
+			continue
+		}
+		addresses := addressesByFamily[string(family)]
+		if len(addresses) == 0 {
+			continue
+		}
+		Plog.Info().Int("ImplIndex", i).Msg("Funding bootstrapped executor transmitters")
+		if err := impl.FundAddresses(ctx, blockchains[i], addresses, big.NewInt(5)); err != nil {
+			return fmt.Errorf("failed to fund bootstrapped executor transmitters: %w", err)
+		}
+		Plog.Info().Int("ImplIndex", i).Msg("Funded bootstrapped executor transmitters")
+	}
+	return nil
 }
 
 // launchExecutors starts executor containers for all Standalone-mode inputs.
--- a/build/devenv/environment.go
+++ b/build/devenv/environment.go
@@ -537,11 +537,9 @@ func buildEnvironmentTopology(in *Cfg, e *deployment.Environment) *ccvdeployment
 // It returns a map of container name -> job spec.
 // The ds parameter is a mutable datastore that will be updated with the changeset output.
 func generateExecutorJobSpecs(
-	ctx context.Context,
 	e *deployment.Environment,
 	in *Cfg,
 	selectors []uint64,
-	impls []cciptestinterfaces.CCIP17Configuration,
 	topology *ccvdeployment.EnvironmentTopology,
 	ds datastore.MutableDataStore,
 ) (map[string]bootstrap.JobSpec, error) {
@@ -606,63 +604,6 @@ func generateExecutorJobSpecs(
 				executorJobSpecs[exec.ContainerName] = executorSpec.ToBootstrapJobSpec()
 			}
 		}
-	}
-
-	// Set transmitter keys using family-specific key generation
-	for _, exec := range in.Executor {
-		family := exec.ChainFamily
-		if family == "" {
-			family = chainsel.FamilyEVM
-		}
-		fac, facErr := GetImplFactory(family)
-		if facErr != nil {
-			return nil, fmt.Errorf("no impl factory for executor chain family %q: %w", family, facErr)
-		}
-		pk, pkErr := fac.GenerateTransmitterKey()
-		if pkErr != nil {
-			return nil, fmt.Errorf("failed to generate transmitter key for family %q: %w", family, pkErr)
-		}
-		exec.TransmitterPrivateKey = pk
-	}
-
-	// Build executor transmitter addresses grouped by chain family so each chain
-	// only funds addresses in its native format.
-	addressesByFamily := make(map[string][]protocol.UnknownAddress)
-	for _, exec := range in.Executor {
-		family := exec.ChainFamily
-		if family == "" {
-			family = chainsel.FamilyEVM
-		}
-		fac, facErr := GetImplFactory(family)
-		if facErr != nil {
-			return nil, fmt.Errorf("no impl factory for executor chain family %q: %w", family, facErr)
-		}
-		addressesByFamily[family] = append(
-			addressesByFamily[family],
-			exec.GetTransmitterAddress(fac.TransmitterAddress),
-		)
-	}
-
-	Plog.Info().Any("AddressesByFamily", addressesByFamily).Int("ImplsLen", len(impls)).Msg("Funding executors")
-	for i, impl := range impls {
-		family, famErr := blockchain.TypeToFamily(in.Blockchains[i].Type)
-		if famErr != nil {
-			continue
-		}
-		fac, facErr := GetImplFactory(string(family))
-		if facErr != nil || !fac.SupportsFunding() {
-			continue
-		}
-		addresses := addressesByFamily[string(family)]
-		if len(addresses) == 0 {
-			continue
-		}
-
-		Plog.Info().Int("ImplIndex", i).Msg("Funding executor")
-		if err := impl.FundAddresses(ctx, in.Blockchains[i], addresses, big.NewInt(5)); err != nil {
-			return nil, fmt.Errorf("failed to fund addresses for executors: %w", err)
-		}
-		Plog.Info().Int("ImplIndex", i).Msg("Funded executors")
 	}
 
 	return executorJobSpecs, nil
@@ -1409,7 +1350,7 @@ func NewEnvironment() (in *Cfg, err error) {
 	// START: Launch executors //
 	/////////////////////////////
 
-	executorJobSpecs, err := generateExecutorJobSpecs(ctx, e, in, selectors, impls, topology, ds)
+	executorJobSpecs, err := generateExecutorJobSpecs(e, in, selectors, topology, ds)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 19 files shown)


integration/pkg/contracttransmitter/keystore_evm_contract_transmitter.go
```diff
@@ -0,0 +1,179 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("keystore sign failed: %w", err)
+			}
+			return tx.WithSignature(signer, signResp.Signature)
+		},
+	}
+
+	auth.Nonce = big.NewInt(int64(nonce)) //nolint:gosec // G115 will replace with txm
+	auth.GasPrice = gasPrice
+	auth.GasLimit = uint64(10000000) // TODO: use a proper gas limit
```
winder, line 117: This is copied over from the RPC based contract transmitter and remains a TODO.

build/devenv/environment.go
```diff
@@ -607,63 +606,6 @@ func generateExecutorJobSpecs(
 		}
 	}
 
-	// Set transmitter keys using family-specific key generation
```
winder, line 610: no longer needed because keystore is being used.

build/devenv/environment.go
```diff
@@ -1785,6 +1731,58 @@ func launchCLNodes(
...
+			break
+		}
+		family, famErr := blockchain.TypeToFamily(blockchains[i].Type)
+		if famErr != nil {
+			continue
+		}
+		fac, facErr := GetImplFactory(string(family))
+		if facErr != nil || !fac.SupportsFunding() {
+			continue
+		}
```
makramkd, line 1772: **note**: This feels a bit awkward, I wonder if this should be on the `Impl` interface instead

build/devenv/environment.go
```diff
@@ -1785,6 +1731,58 @@ func launchCLNodes(
...
+		addresses := addressesByFamily[string(family)]
+		if len(addresses) == 0 {
+			continue
+		}
+		Plog.Info().Int("ImplIndex", i).Msg("Funding executor transmitters")
+		if err := impl.FundAddresses(ctx, blockchains[i], addresses, big.NewInt(5)); err != nil {
```
makramkd, line 1778: **note**: the `big.NewInt(5)` here is interesting, I think the assumption is that its in Ether rather than Wei, SOL instead of Lamport, etc.? Looking at the documentation for the method this particular detail isn't documented at all.

build/devenv/services/bootstrap.go
```diff
@@ -97,21 +98,14 @@ type BootstrapKeys struct {
 	ECDSAPublicKey string `toml:"ecdsa_public_key,omitempty"`
 	// ECDSAAddress is the Ethereum address derived from the ECDSA public key (verifier only).
 	ECDSAAddress string `toml:"ecdsa_address,omitempty"`
+	// EVMTransmitterAddress is the Ethereum address derived from the executor's EVM transmitter key.
+	// Only populated for executor nodes.
+	EVMTransmitterAddress string `toml:"evm_transmitter_address,omitempty"`
```
makramkd, line 103: **question**: what is the plan for non-EVM executors? Ideally we can avoid updating this struct if a new family is building their executor.

integration/pkg/accessors/evm/factory.go
```diff
@@ -134,7 +133,14 @@ func (f *factory) GetAccessor(ctx context.Context, chainSelector protocol.ChainS
 	}
 
 	var evmDestReader chainaccess.DestinationReader
-	if destCfg, ok := f.destChainConfigs[chainSelector]; ok && destCfg.OffRampAddress != "" {
+	var offRampAddr common.Address
```
makramkd, line 141: **nit**: declare this closer to the `if` statement where it is updated.

integration/pkg/accessors/evm/factory.go
```diff
@@ -151,50 +157,68 @@ func (f *factory) GetAccessor(ctx context.Context, chainSelector protocol.ChainS
...
-		rpcURL,
-		f.transmitterPrivateKey,
-		common.HexToAddress(destCfg.OffRampAddress),
-	)
-	if err != nil {
-		f.lggr.Warnw("Failed to create EVM contract transmitter, ContractTransmitter will be unavailable", "chainSelector", chainSelector, "error", err)
-		return nil
-	}
-	return ct
+	rpcURL := f.rpcURLs[chainSelector]
```
makramkd, line 165: **issue**: `offRampAddr` isn't checked to be nonzero after the previous `if destCfg.OffRampAddress != ""` statement.

pkg/chainaccess/interfaces.go
```diff
@@ -82,6 +82,17 @@ type Accessor interface {
...
+	// Optional hook: bootstrap.KeystoreSetter
+	//
+	// Accessors that require a keystore for signing (e.g. to build a keystore-backed
+	// ContractTransmitter) should also implement bootstrap.KeystoreSetter:
+	//
+	//   SetKeystore(ks keystore.Keystore)
+	//
+	// bootstrap.KeystoreRegistry checks for this interface after every GetAccessor call
+	// and invokes SetKeystore automatically, so implementations do not need to arrange
+	// injection themselves.
```
makramkd, line 95: **nit**: I think this should be placed in the go doc struct comment so that it shows up when people hover over the `Accessor` interface in their IDE or they use the `go doc` tool to look up the docstring.

executor/const.go
```diff
@@ -1,3 +1,13 @@
...
+const (
+	DefaultConfigFile = "/etc/config.toml"
+
+	// DefaultEVMTransmitterKeyName is the full keystore path of the ECDSA key used
+	// to sign and submit OffRamp transactions on EVM chains. The evm/tx/ prefix
+	// follows the evmkeys convention so that TxKey can use it directly with
+	// WithNoPrefix(). The bootstrap framework creates or loads it on startup;
+	// devenv uses it to look up the funded on-chain address after the container
+	// is running.
+	DefaultEVMTransmitterKeyName = "evm/tx/executor_evm_transmitter_key"
```
makramkd, line 12: **issue**: I don't think we should put EVM-specific key names in the chain-agnostic `executor` package. Can we move instead to pkg/accessors/evm?

integration/pkg/contracttransmitter/keystore_evm_contract_transmitter.go
```diff
@@ -0,0 +1,146 @@
...
+var _ chainaccess.ContractTransmitter = &KeystoreEVMContractTransmitter{}
+
+// KeystoreEVMContractTransmitter submits OffRamp execute transactions using a
+// keystore-managed secp256k1 key. Private key material never leaves the
+// keystore; signing is delegated via [evmkeys.TxKey.GetTransactOpts].
+//
+// This is the keystore-backed equivalent of [EVMContractTransmitter].
+// The TxManager-backed [TXMEVMContractTransmitter] is preferred for production
+// because it handles retries, gas bumping, and nonce management; this
+// transmitter is suitable for simpler deployments or testing.
```
makramkd, line 34: **question**: are we planning on implementing the `TXMEVMContractTransmitter` using the keystore in a followup?

pkg/chainaccess/registry.go
```diff
@@ -148,6 +155,10 @@ type DestinationChainConfig struct {
 	OffRampAddress string `toml:"off_ramp_address"`
 	// RmnAddress is the address of the RMN Remote contract on the destination chain.
 	RmnAddress string `toml:"rmn_address"`
+	// TransmitterKeyName is the name of the ECDSA key in the keystore used to sign and submit
+	// transactions to the OffRamp on this chain. If empty, the EVM accessor defaults to
+	// executor.DefaultEVMTransmitterKeyName.
+	TransmitterKeyName string `toml:"transmitter_key_name"`
```
makramkd, line 161: **question**: does this need to be in the config? I feel like specifying key name in either config or in the `bootstrap.Run` call may be confusing. Use case for config-driven would be if we want to change the transmitter key w/out re-building the app, for example.

integration/pkg/contracttransmitter/keystore_evm_contract_transmitter.go
```diff
@@ -0,0 +1,146 @@
...
+		chainID:       chainID,
+		Client:        client,
+		OffRamp:       *boundContract,
+		chainSelector: chainSelector,
+	}, nil
+}
+
+// GetTransactOpts builds [bind.TransactOpts] backed by the keystore key.
+// Nonce and gas price are fetched live from the pending state.
+func (ct *KeystoreEVMContractTransmitter) GetTransactOpts(ctx context.Context) (*bind.TransactOpts, error) {
```
makramkd: **nit**: this can be internal/private.


```suggestion
func (ct *KeystoreEVMContractTransmitter) getTransactOpts(ctx context.Context) (*bind.TransactOpts, error) {
```

integration/pkg/contracttransmitter/keystore_evm_contract_transmitter.go
```diff
@@ -0,0 +1,146 @@
...
+	chainID.SetString(id, 10)
+
+	txKeys, err := evmkeys.GetTxKeys(ctx, ks, []string{keyName}, evmkeys.WithNoPrefix())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get tx key %q from keystore: %w", keyName, err)
+	}
+	if len(txKeys) == 0 {
+		return nil, fmt.Errorf("key %q not found in keystore", keyName)
+	}
+
```
makramkd, line 81: **nit**: sanity check with a `if len(txKeys) == 1`? Or perhaps simplify by making the only check `if len(txKeys) != 1`.

integration/pkg/contracttransmitter/evm_contract_transmitter.go
makramkd, line 1: **question**: can this file be deleted entirely? I don't think `NewEVMContractTransmitterFromRPC` will be used after this point?

integration/pkg/accessors/evm/factory.go
```diff
@@ -151,50 +157,68 @@ func (f *factory) GetAccessor(ctx context.Context, chainSelector protocol.ChainS
...
 		destinationReader:   destinationReader,
 		contractTransmitter: contractTransmitter,
 	}
 }
 
+// SetKeystore builds and installs a keystore-backed ContractTransmitter, replacing any
+// previously set transmitter (e.g. the env-var fallback). No-op when ks is nil,
+// keyName is empty, or no RPC URL is available for this chain.
+func (a *accessor) SetKeystore(ks keystore.Keystore) {
+	if ks == nil || a.keyName == "" || a.rpcURL == "" {
```
makramkd: **question**: Could the lack of logging here potentially cause confusion? I wonder if a single `a.lggr.Warnw` indicating that the keystore won't be set due to missing dependencies could potentially help if someone were debugging. I believe that `keyName` and `rpcURL` not being set is basically a misconfiguration right?

build/devenv/environment.go
winder: Yeah, `SupportsFunding()` was one of the functions @tt-cll added while trying to organize cross-family dependenciese. They landed in `ImplFactory` and we all agreed that it wasn't the best spot, but was the best we could do at the time. All of them (including this) need to move at some point.

build/devenv/services/bootstrap.go
winder: I'm not sure. The `BootstrapKeys` struct is a cross-service integration point that I'd like to get rid of. I'm not sure how to do that yet. It probably requires thinking through the ephemeral devenv phases through a bit more, my initial proposal didn't consider funding addresses.

winder: I added a note to the design doc about funding addresses.

integration/pkg/accessors/evm/factory.go
winder: Looks like onRampAddresse and rmnRemoteAddress has the same problem.

pkg/chainaccess/interfaces.go
winder: On one hand, I definitely agree. Autocomplete is something I was considering when adding this comment. The problem is that I didn't want to add the keystore dependency to this package.

If we're going to expand the accessor to include optional "test" interfaces as well, we'll have to find a way to manage this type of optional dependency.

executor/const.go
winder: ~Good call~ Actually, I'm not sure about making a new `evm` package, it moves us away from injecting family specific behavior.

I actually wanted to define this in the main.go function (and/or the config file) so that it didn't have any scope or dependencies.

integration/pkg/contracttransmitter/keystore_evm_contract_transmitter.go
winder: We already have a TXMEVMContractTransmitter, it's in `txm_evm_contract_transmitter.go`

pkg/chainaccess/registry.go
winder: I went back and forth on this point. It depends on how we recommend NOPs manage their keystore. A name must be assigned and it felt weird to have the names be hard coded.

i.e. suppose there was a key rotation. Without a config, you have to override the existing key and it takes place immediately. I would personally prefer to configure keystore with both keys and swap in the new one by updating a config. The config approach supports both patterns.

integration/pkg/contracttransmitter/evm_contract_transmitter.go
winder: It probably wont be used. I don't personally mind keeping it around for reference. We'll have to revisit these files anyway when migrating stuff to a separate EVM module.

integration/pkg/accessors/evm/factory.go
winder: Good call. I added a log to the keystore registry for exactly this reason but missed this point.

integration/pkg/contracttransmitter/keystore_evm_contract_transmitter.go
winder: Don't you think multiple keys should be supported? Maybe this is a warning that only one of the configured keys is being used?

makramkd: Eventually we might support multiple keys, but for the code as-is, which passes just a single `keyName`, it should only ever return 1 key.

integration/pkg/accessors/evm/factory.go
makramkd: Should we create a ticket for that? It seems odd to be able to get to this point with a misconfiguration like that.

executor/const.go
makramkd: I don't think thats a new package, its the one that has e.g. `integration/pkg/contracttransmitter/evm_contract_transmitter.go`

makramkd (approved):

integration/pkg/accessors/evm/factory.go
winder: I added a check to all of them, we should be good now.

executor/const.go
winder: I'll consider this for a future PR. The suggestion would be a similar sort of convergence issue we've been discussing with the devenv. The const goes away entirely in a future version of the devenv where we have the funding issue is sorted out.

tt-cll (approved):

PR #1060, opened 2026-04-29
add/remove nop from committee

tt-cll:


deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/smartcontractkit/chainlink-ccv/deployment/adapters"
+	"github.com/smartcontractkit/chainlink-ccv/deployment/operations/fetch_signing_keys"
+)
+
+// AddNOPToCommitteeInput is the input for step-1 of the AddNOP two-entry product.
+type AddNOPToCommitteeInput struct {
+	// CommitteeQualifier identifies the committee the NOP is joining.
+	CommitteeQualifier string
+	// ChainSelectors are the destination chains on which the committee verifier is deployed.
+	ChainSelectors []uint64
```
carte7000: Question: In practise the quorum config is based on source chain. Ideally when we add a NOP to a committee for a chain we automatically replicate it to all destination.

By sending the ChainSelectors in the params we open the door to having a partial state replication. That might be intentional here but just flagging potential issue.

Maybe what we want this CS to do is say add NOP to committee on specific source chains and then under the hood it replicate to all connected dest?

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,228 @@
...
+
+	return deployment.CreateChangeSet(apply, validate)
+}
+
+// AddNOPOffchain is step-2 of the AddNOP two-entry product.
+//
+// It regenerates the aggregator config from the updated onchain state (which now includes
+// the new signer) and writes it to the DataStore for each listed service identifier.
+// Triggered by the CLD post-proposal hook after timelock execution.
+func AddNOPOffchain(registry *adapters.Registry) deployment.ChangeSetV2[AddNOPOffchainInput] {
```
carte7000, line 220: Do we want this to update jobs as well? Or we would keep the ApplyVerifier CS for this part?

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,228 @@
...
+			}
+			if a.Aggregator == nil {
+				return fmt.Errorf("chain %d: no Aggregator adapter registered", sel)
+			}
+		}
+		return nil
+	}
+
+	apply := func(e deployment.Environment, cfg AddNOPOffchainInput) (deployment.ChangesetOutput, error) {
+		committee, err := buildAggregatorCommittee(e, registry, cfg.CommitteeQualifier, cfg.ChainSelectors, nil)
```
carte7000: Sending a subset of the env selector to this function might remove chains that are out of scope of this CS from the final aggregator config. IIRC the output of this function will be the entire aggregator config, so if we omitted a specific chain in cfg.ChainSelectors we might get a new aggregator config that exclude that chain. If we deploy this the aggregator will stop processing message from that omitted chain

deployment/changesets/add_nop_to_committee.go
tt-cll: Yes we do. Added

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,457 @@
...
+		return nil
+	}
+
+	apply := func(e deployment.Environment, cfg AddNOPOffchainInput) (deployment.ChangesetOutput, error) {
+		destChains := registry.AllDeployedCommitteeVerifierChains(e.DataStore, cfg.CommitteeQualifier)
+		if len(destChains) == 0 {
+			return deployment.ChangesetOutput{}, fmt.Errorf("no dest chains found for committee %q", cfg.CommitteeQualifier)
+		}
+
+		committee, err := buildAggregatorCommittee(e, registry, cfg.CommitteeQualifier, destChains, nil)
```
carte7000: nit: I was caught offguard with the destChains name here. This seems to be the entire set of chains that this committee is on right? The aggregator needs the entire set of chains

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,457 @@
...
+	signerAddress := cfg.ExpectedSignerAddress
+	if signerAddress == "" {
+		addr, err := fetchSignerAddress(e, string(cfg.NOPAlias), signerFamily)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to fetch signer address for job provisioning: %w", err)
+		}
+		signerAddress = addr
+	}
+
+	contractAddresses, err := buildVerifierContractConfigs(registry, e, destChains, cfg.CommitteeQualifier, cfg.ExecutorQualifier)
```
carte7000: dito here: The verifier need the entire set of chains for the committee

carte7000 (dismissed):

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,457 @@
...
+	// The changeset automatically updates every destination chain that has this committee
+	// verifier deployed — callers do not enumerate dest chains.
+	SourceChainSelectors []uint64
+	// NOPAlias is the node alias used to look up the NOP's signing address in JD.
+	NOPAlias string
+	// NewThreshold is the desired threshold after the NOP is added. Zero keeps the current threshold.
+	NewThreshold uint8
+}
+
+// AddNOPOffchainInput is the input for step-2 of the AddNOP two-entry product.
```
makramkd, line 70: It's good that we have the step 1 and step 2 documented in detail - just wondering if we should have a go doc string somewhere else in this file that describes the entire two-step process holistically.

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,457 @@
...
+	// DisableFinalityCheckers is a list of chain selectors for which finality checks are disabled.
+	DisableFinalityCheckers []string
+	// Monitoring holds monitoring configuration included in the job spec.
+	Monitoring ccvdeployment.MonitoringConfig
+}
+
+// AddNOPToCommittee is step-1 of a coupled onchain-first two-entry product.
+//
+// It fetches the NOP's signing address from JD, then for every destination chain that has
+// this committee verifier deployed (discovered automatically from the registry) it reads the
```
makramkd, line 99: Wanted to point out that the "fetch NOP signing address from JD" probably will not work in standalone mode (the mode that's not using the Chainlink node). Is there a way to support this compound changeset without JD? If the answer is no, this implies that standalone mode would have to behave similar to the Chainlink node, where it would send NOP keys to JD, which is work that isn't scheduled yet.

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,457 @@
...
+// current onchain committee state and submits an applySignatureConfigs call that appends the
+// new signer to each of the specified source chain configs. The aggregator config regen is
+// deferred to AddNOPOffchain (step-2), which runs after the timelock executes via the CLD
+// post-proposal hook.
```
makramkd, line 103: How would post-proposal hooks work outside of the CLD context, with e.g. devenv?

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,457 @@
...
+	e deployment.Environment,
+	registry *adapters.Registry,
+	committeeQualifier string,
+	nopAlias string,
+	signerFamily string,
+	sourceChainSelectors []uint64,
+	newThreshold uint8,
+	buildChange func(*adapters.CommitteeState, string, uint8, []uint64) (adapters.SignatureConfigChange, error),
+) error {
+	ctx := context.Background()
```
makramkd: Why this vs. e.g. `e.GetContext()`?

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,457 @@
...
+		if err != nil {
+			return fmt.Errorf("dest chain %d: %w", sel, err)
+		}
+		if len(change.NewConfigs) == 0 {
+			continue // this dest chain has no configs for the requested source chains
+		}
+		a, _ := registry.GetByChain(sel)
+		if err := a.CommitteeVerifierOnchain.ApplySignatureConfigs(ctx, e, sel, committeeQualifier, change); err != nil {
+			return fmt.Errorf("dest chain %d: ApplySignatureConfigs failed: %w", sel, err)
+		}
```
makramkd, line 193: `err` is double-checked here w/out actually changing.

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,457 @@
...
+			}
+			if a.Aggregator == nil {
+				return fmt.Errorf("dest chain %d: no Aggregator adapter registered", sel)
+			}
+		}
+
+		// Safety backstop: assert the new signer is present onchain on every dest chain for
+		// every source chain. Catches hook misfires and out-of-order manual invocations.
+		if cfg.ExpectedSignerAddress != "" {
+			ctx := context.Background()
```
makramkd: Ditto here, `e.GetContext()`? Not sure who actually injects this context tbh.

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,457 @@
...
+
+	contractAddresses, err := buildVerifierContractConfigs(registry, e, committeeChains, cfg.CommitteeQualifier, cfg.ExecutorQualifier)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to build verifier contract configs: %w", err)
+	}
+
+	nopInputs := []verifierNOPInput{{
+		Alias:                 cfg.NOPAlias,
+		SignerAddressByFamily: map[string]string{signerFamily: signerAddress},
+		Mode:                  shared.NOPModeCL,
```
makramkd, line 360: This is another place that seems to be CL-mode only, I think in a follow-up we should think about what supporting standalone would look like.

deployment/changesets/add_nop_to_committee.go
```diff
@@ -0,0 +1,457 @@
...
+	manageReport, err := operations.ExecuteSequence(
+		e.OperationsBundle,
+		sequences.ManageJobProposals,
+		sequences.ManageJobProposalsDeps{Env: envWithBase},
+		sequences.ManageJobProposalsInput{
+			JobSpecs:      jobSpecs,
+			AffectedScope: scope,
+			Labels: map[string]string{
+				"job_type":  "verifier",
+				"committee": cfg.CommitteeQualifier,
```
makramkd: Small nit but maybe these keys can be constants, if we expect to use them in other job proposal type operations.

deployment/changesets/remove_nop_from_committee.go
```diff
@@ -0,0 +1,316 @@
...
+			}
+			if a.Aggregator == nil {
+				return fmt.Errorf("dest chain %d: no Aggregator adapter registered", sel)
+			}
+		}
+
+		// Safety backstop: assert the removed signer is absent onchain on every dest chain for
+		// every source chain. Catches hook misfires and out-of-order manual invocations.
+		if cfg.RemovedSignerAddress != "" {
+			ctx := context.Background()
```
makramkd: Ditto here re: `e.GetContext()`

deployment/changesets/add_nop_to_committee.go
tt-cll: I don't know of a good way to test that flow. Devenv's orchestration can just do everything in a single call since it's not distributed/time delayed like a live env

deployment/changesets/add_nop_to_committee.go
tt-cll: Opened https://smartcontract-it.atlassian.net/browse/CCIP-11342 to track changesets for standalone

makramkd (approved):

carte7000 (approved):

PR #1061, opened 2026-04-29
bootstrap: allow explicit key declarations via WithKey

winder:
## Description

Removes the hardcoded default key list from initializeKeystore. Services now declare the keys they
  need via WithKey options passed to bootstrap.Run, making key requirements explicit and visible at the
   call site. The bootstrapper has no policy of its own — it only initialises what it is told to.

## Testing

This feature is non-functional refactoring that replaces code paths hit by all existing integration tests. If the environment starts and messages execute, the changes were a success.

## Checklist

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -297,13 +297,20 @@ func initializeKeystore(ctx context.Context, lggr logger.Logger, db *sqlx.DB, ks
 		return nil, nil, fmt.Errorf("failed to load keystore: %w", err)
 	}
 
+	var csaKeyName string
 	for _, k := range requiredKeys {
 		if err := keys.EnsureKey(ctx, lggr, ks, k.name, k.purpose, k.keyType); err != nil {
-			return nil, nil, fmt.Errorf("failed to ensure %s key: %w", k.purpose, err)
+			return nil, nil, fmt.Errorf("failed to ensure key %q (purpose=%q, type=%v): %w", k.name, k.purpose, k.keyType, err)
+		}
+		if k.purpose == "csa" {
+			csaKeyName = k.name
 		}
 	}
+	if csaKeyName == "" {
+		return nil, nil, fmt.Errorf("no key with purpose %q declared; a CSA key is required for JD communication", "csa")
+	}
 
-	csaSigner, err := keys.NewCSASigner(ctx, ks, keys.DefaultCSAKeyName)
+	csaSigner, err := keys.NewCSASigner(ctx, ks, csaKeyName)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to get csa signer: %w", err)
 	}
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -288,12 +288,7 @@ func initializeKeystore(ctx context.Context, lggr logger.Logger, db *sqlx.DB, ks
 		return nil, nil, fmt.Errorf("failed to load keystore: %w", err)
 	}
 
-	defaultKeys := []keyToInit{
-		{keys.DefaultCSAKeyName, "csa", keystore.Ed25519},
-		{keys.DefaultECDSASigningKeyName, "signing", keystore.ECDSA_S256},
-		{keys.DefaultEdDSASigningKeyName, "signing", keystore.Ed25519},
-	}
-	for _, k := range append(defaultKeys, requiredKeys...) {
+	for _, k := range requiredKeys {
 		if err := keys.EnsureKey(ctx, lggr, ks, k.name, k.purpose, k.keyType); err != nil {
 			return nil, nil, fmt.Errorf("failed to ensure %s key: %w", k.purpose, err)
 		}
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -94,6 +94,7 @@ type Bootstrapper struct {
 	config           *Config
 	lifecycleManager *lifecycle.Manager
 	infoServer       *infoServer
+	keys             []keyToInit
 
 	// application
 	appCfg *string
@@ -180,7 +181,7 @@ func (b *Bootstrapper) startWithJDLifecycle(ctx context.Context) error {
 		return fmt.Errorf("failed to connect to bootstrapper database: %w", err)
 	}
 
-	keyStore, csaSigner, err := initializeKeystore(ctx, b.lggr, db, b.config.Keystore.Password)
+	keyStore, csaSigner, err := initializeKeystore(ctx, b.lggr, db, b.config.Keystore.Password, b.keys)
 	if err != nil {
 		return fmt.Errorf("failed to initialize keystore: %w", err)
 	}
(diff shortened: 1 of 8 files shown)


cmd/executor/standalone/main.go
```diff
@@ -5,14 +5,18 @@ import (
...
 	_ "github.com/smartcontractkit/chainlink-ccv/integration/pkg/accessors/evm" // evm accessor driver
+	"github.com/smartcontractkit/chainlink-common/keystore"
 )
 
 func main() {
 	err := bootstrap.Run(
 		"Executor",
 		cmdexecutor.NewFactory(),
+		bootstrap.WithKey(keys.DefaultCSAKeyName, "csa", keystore.Ed25519),   // node identity key for JD communication
+		bootstrap.WithKey("default_evm_key", "signing", keystore.ECDSA_S256), // EVM signing key for OffRamp transaction submission
```
winder: These are now specified by the app rather than hard coded defaults.

bootstrap/bootstrap.go
```diff
@@ -122,6 +123,15 @@ func NewBootstrapper(
...
 	}
 
+	// Backwards compatibility: if no keys are declared, initialize the original default set.
+	if len(b.keys) == 0 {
+		b.keys = []keyToInit{
+			{keys.DefaultCSAKeyName, "csa", keystore.Ed25519},
+			{keys.DefaultECDSASigningKeyName, "signing", keystore.ECDSA_S256},
+			{keys.DefaultEdDSASigningKeyName, "signing", keystore.Ed25519},
+		}
+	}
```
winder, line 134: Hard coded defaults remain for backwards compatibility.

cmd/verifier/committee/main.go
```diff
@@ -22,6 +24,9 @@ func main() {
 		"EVMCommitteeVerifier",
 		cmd.NewCommitteeVerifierServiceFactory(),
 		bootstrap.WithLogLevel(zapcore.InfoLevel),
+		bootstrap.WithKey(keys.DefaultCSAKeyName, "csa", keystore.Ed25519),                 // node identity key for JD communication
```
winder: I'm considering moving this into the bootstrap service and having it defined automatically when JD mode is used.

bootstrap/bootstrap.go
makramkd: **question**: Are we thinking to axe these at some point? Wondering if we should mark as "Deprecated".

bootstrap/bootstrap.go
```diff
@@ -321,6 +329,28 @@ func WithLogLevel(logLevel zapcore.Level) Option {
 	}
 }
 
+type keyToInit struct {
+	name    string
+	purpose string
```
makramkd, line 350: **question**: Should we have constants for these purposes?

cmd/verifier/committee/main.go
```diff
@@ -22,6 +24,9 @@ func main() {
...
+		bootstrap.WithKey(keys.DefaultECDSASigningKeyName, "signing", keystore.ECDSA_S256), // ECDSA key for signing verification results
```
winder, line 27: This seems to be dead code.

bootstrap/bootstrap.go
```diff
@@ -321,6 +329,28 @@ func WithLogLevel(logLevel zapcore.Level) Option {
...
+	keyType keystore.KeyType
+}
+
+// WithKey declares a key that the bootstrapper must ensure exists, creating it if absent.
+// When no WithKey options are provided, the bootstrapper applies a default set of three keys:
+// keys.DefaultCSAKeyName (Ed25519), keys.DefaultECDSASigningKeyName (ECDSA_S256), and
+// keys.DefaultEdDSASigningKeyName (Ed25519). Passing one or more WithKey options suppresses
+// those defaults entirely; the caller is responsible for declaring every key it requires.
+func WithKey(name, purpose string, keyType keystore.KeyType) Option {
```
makramkd, line 359: **question**: Should we also document that the CSA key is needed if WithJD() is used?

build/devenv/services/bootstrap.go
```diff
@@ -98,19 +98,22 @@ type BootstrapKeys struct {
...
+func GetExecutorBootstrapKeys(bootstrapURL string) (BootstrapKeys, error) {
+	return fetchBootstrapKeys(bootstrapURL, []string{bskeys.DefaultCSAKeyName})
+}
+
+// GetBootstrapKeys fetches the CSA and ECDSA keys from the bootstrap server.
+// Verifiers need both for JD registration and signing.
+func GetBootstrapKeys(bootstrapURL string) (BootstrapKeys, error) {
+	return fetchBootstrapKeys(bootstrapURL, []string{bskeys.DefaultCSAKeyName, bskeys.DefaultECDSASigningKeyName})
+}
+
```
winder, line 113: Are these needed? This logic couples the hard coded app config to the devenv so that they can be copied into the output file.

cmd/executor/standalone/main.go
makramkd: **question**: while it is true that this key is signing transactions, it also requires funds to transmit them. Should the key purpose still be "signing" or "transmitting" in this case?

cmd/verifier/committee/main.go
```diff
@@ -22,6 +24,9 @@ func main() {
...
+		bootstrap.WithKey(keys.DefaultEdDSASigningKeyName, "signing", keystore.Ed25519),    // EdDSA key for signing verification results
```
makramkd: **suggestion**: I think this should be removed to allay confusion, since the EVM verifier only uses ECDSA.

deployment/adapters/registry.go
```diff
@@ -11,11 +11,11 @@ import (
 // ChainAdapters bundles all chain-family-specific adapter implementations.
 // A nil field means the adapter is not supported for that family.
 type ChainAdapters struct {
-	Aggregator              AggregatorConfigAdapter
```
makramkd, line 14: @tt-cll is it possible we aren't running linting/formatting etc. on the new `deployment` mod?

cmd/verifier/committee/main.go
makramkd: I was thinking the same thing, either that or document that it must be provided when `WithJD` is used

cmd/executor/standalone/main.go
winder: Good call, I was mostly fooling around with the interface here. Need to revisit the actual key usage in a followup.

build/devenv/services/bootstrap.go
makramkd: I believe I added these initially just to confirm the keystore stuff was working E2E, I think they can be axed.

cmd/verifier/committee/main.go
winder: Totally agree. This was a mystery to me before and I didn't realize we had unused stuff.

bootstrap/bootstrap.go
winder: I'm actually thinking that all of these consts can go away:
* `DefaultEdDSASigningKeyName` - scoped to the committee verifier main.go file and passed into the `NewCommitteeService` function.
* `DefaultCSAKeyName` - a private const scoped to the bootstrapper.
* `DefaultECDSASigningKeyName` - deleted, it is unused anywhere.

deployment/adapters/registry.go
tt-cll: It does look like the lint workflow manually lists out every directory to lint, and deployment was not added there. https://github.com/smartcontractkit/chainlink-ccv/blob/main/.github/workflows/golangci-lint.yaml

We ran into an issue last week where the linter in the PR ran `just lint` but the one in the merge queue ran with a different config. It seems more intuitive to have `just lint` be the all in one that is run locally and from CI so there's not a chance of a discrepancy.

makramkd: I can create an issue for that

bootstrap/bootstrap.go
makramkd: > DefaultEdDSASigningKeyName - scoped to the committee verifier main.go file and passed into the NewCommitteeService function.
> DefaultECDSASigningKeyName - deleted, it is unused anywhere.

Just confirming, I think you mean the other way around, right? ECDSA is currently used, EdDSA not

build/devenv/services/bootstrap.go
winder: Turns out they're very important. They end up in the output which is eventually used to copy public keys into config files for JD + aggregator.

makramkd (approved):

carte7000 (approved):

bootstrap/bootstrap.go
winder: Yes, I got this right in the implementation.

PR #1065, opened 2026-04-30
fix token pool pairs

tt-cll:
## Description

Fixes generic token transfer setup for mixed EVM/Canton environments.

The previous token-pool qualifiers were directional, so the same logical token pair could produce different local pool refs depending on which side built the config. That made batching brittle: reciprocal EVM -> Canton and Canton -> EVM `TokenTransferConfig`s could be split apart even though `ConfigureTokensForTransfers` expects both sides of a remote chain relationship in the same call.

This PR makes token-pool qualifiers reciprocal by separating the logical pair identity from the concrete pool instance, then updates batching to group configs by the canonical pair while preserving each pool’s full `RemoteChains` set. It also adds validation so missing reciprocal pool configs fail earlier with a useful error instead of surfacing as harder-to-debug pool compatibility failures.

Related Canton consumer PR: https://github.com/smartcontractkit/chainlink-canton/pull/440

## Testing

devenv tests

## Checklist

### DIFF
--- a/build/devenv/implcommon_test.go
+++ b/build/devenv/implcommon_test.go
@@ -16,12 +16,13 @@ func TestBuildTokenTransferBatchesMultipleTokensSameSelectors(t *testing.T) {
 	aBar := testTokenPoolRef(1, "bar")
 	bBar := testTokenPoolRef(2, "bar")
 
-	batches := buildTokenTransferBatches([]tokenscore.TokenTransferConfig{
+	batches, err := buildTokenTransferBatches([]tokenscore.TokenTransferConfig{
 		testTokenTransferConfig(aFoo, bFoo),
 		testTokenTransferConfig(bFoo, aFoo),
 		testTokenTransferConfig(aBar, bBar),
 		testTokenTransferConfig(bBar, aBar),
 	})
+	require.NoError(t, err)
 	require.Len(t, batches, 2)
 
 	for _, batch := range batches {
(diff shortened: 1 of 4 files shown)


build/devenv/implcommon_test.go
```diff
@@ -87,6 +87,34 @@ func TestBuildTokenTransferBatchesAsymmetricPoolsSplitByLocalPool(t *testing.T)
 	require.Equal(t, map[string]bool{"burn": true, "lock": true}, seenQualifiers)
 }
 
+// TestBuildTokenTransferBatchesCrossTypePools verifies that an EVM
+// BurnMintTokenPool and a Canton LockReleaseTokenPool for the same token pair
+// land in the same batch. Both carry the canonical pair qualifier produced by
+// TokenCombination, differing only in the ":local" / ":remote" suffix.
+func TestBuildTokenTransferBatchesCrossTypePools(t *testing.T) {
```
makramkd, line 94: Are these tests running in CI btw?

makramkd (dismissed):

carte7000 (dismissed):

build/devenv/implcommon_test.go
tt-cll: hmmm I don't think so. Will opened https://smartcontract-it.atlassian.net/browse/CCIP-11254 so we can try to enable these running as part of that

build/devenv/implcommon_test.go
```diff
@@ -40,51 +41,201 @@ func TestBuildTokenTransferBatchesSymmetricTokenAcrossAllLanes(t *testing.T) {
 	pool2 := testTokenPoolRef(2, "burn")
 	pool3 := testTokenPoolRef(3, "burn")
 
-	batches := buildTokenTransferBatches([]tokenscore.TokenTransferConfig{
+	batches, err := buildTokenTransferBatches([]tokenscore.TokenTransferConfig{
 		testTokenTransferConfig(pool1, pool2, pool3),
 		testTokenTransferConfig(pool2, pool1, pool3),
 		testTokenTransferConfig(pool3, pool1, pool2),
```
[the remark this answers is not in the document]
tt-cll: not valid in this version of golang

makramkd (approved):
Approving to unblock, but IMO should revisit the algorithm as it seems quite complex.

carte7000 (approved):

PR #1066, opened 2026-05-01
chore: add codeql scanning

KodeyThomas:
Adds in CodeQL scanning, should be the same workflow as used in the core node modified slightly to remove uneeded build steps (for JS/TS UI)


carte7000 (approved):

RensR (approved):
Approved with the note that we will use these from `.github` instead in a future PR

PR #1072, opened 2026-05-04
Phased DevEnv Design Validation

winder:
## Description
This PR should not be merged. I'm validating the approach and looking for problems.

Notes:
* Overall phased startup appears to work.
* The ImplFactory is a cross-component concern which doesn't fit well with the componentized design. It looks like ImplFactory can be split across multiple modules which means it can go away entirely. This will generally simplify the business logic so that it is confined to a specific component rather than a big interface with multiple concerns.
* The "Fallback" approach works well for backwards compatibility.
* When configuration changes are needed, introducing per-component version fields can be used for backwards compatibility.
* The 4 phase directed acyclic graph (DAG) is simplified and may benefit from a more complex approach in the future. Splitting the business logic from DAG generation as done in this prototype makes that a fairly straightforward feature to add in the future.
* A balance needs to be made for what is hidden in components and what is exposed via the runtime.
    * Funding -> how can a component request that a key be funded?
    * The "Out" values are passed along between phases, how are types managed?

## Testing
N/A

## Checklist

### DIFF
--- a/build/devenv/cciptestinterfaces/interface.go
+++ b/build/devenv/cciptestinterfaces/interface.go
@@ -354,10 +354,8 @@ type DeployerNonceBumper interface {
 }
 
 // OffChainConfigurable defines methods that allows to
-// deploy a local blockchain network for tests and configure CL nodes for Chainlink product.
+// configure CL nodes for Chainlink product.
 type OffChainConfigurable interface {
-	// DeployLocalNetwork deploy local node of network X
-	DeployLocalNetwork(ctx context.Context, bcs *blockchain.Input) (*blockchain.Output, error)
 	// ConfigureNodes configure CL nodes from blockchain data
 	// returns a piece of TOML config as a string that the framework inject into final configuration
 	ConfigureNodes(ctx context.Context, blockchain *blockchain.Input) (string, error)
--- a/build/devenv/cli/ccv.go
+++ b/build/devenv/cli/ccv.go
@@ -37,6 +37,9 @@ const (
 	LocalCCVDashboard      = "http://localhost:3000/d/f8a04cef-653f-46d3-86df-87c532300672/ccv-services?orgId=1&refresh=5s"
 )
 
+// newEnvFn is set by PersistentPreRunE based on the --env-mode flag.
+var newEnvFn func() (*ccv.Cfg, error)
+
 var rootCmd = &cobra.Command{
 	Use:   "ccv",
 	Short: "A CCV local environment tool",
@@ -48,6 +51,18 @@ var rootCmd = &cobra.Command{
 		if debug {
 			framework.L.Info().Msg("Debug mode enabled, setting CTF_CLNODE_DLV=true")
 			os.Setenv("CTF_CLNODE_DLV", "true")
+		}
+		mode, err := cmd.Flags().GetString("env-mode")
+		if err != nil {
+			return err
+		}
+		switch mode {
+		case "legacy":
+			newEnvFn = ccv.NewEnvironment
+		case "phased":
+			newEnvFn = ccv.NewEnvironmentPhased
+		default:
+			panic(fmt.Sprintf("unknown --env-mode %q", mode))
 		}
 		return nil
 	},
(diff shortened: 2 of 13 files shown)


build/devenv/runtime/environment.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Phase 1
+	for _, key := range sortedKeys(specific) {
+		comp := specific[key]
+		if p1, ok := comp.(Phase1Component); ok {
+			out, err := p1.RunPhase1(ctx, rawConfig, rawConfig[key])
+			if err != nil {
+				return nil, fmt.Errorf("phase1 %s: %w", key, err)
+			}
+			maps.Copy(accumulated, out)
```
winder, line 51: make sure Copy doesn't overwrite anything

build/devenv/runtime/environment.go
```diff
@@ -0,0 +1,151 @@
...
+			out, err := p1.RunPhase1(ctx, rawConfig, unclaimed)
+			if err != nil {
+				return nil, fmt.Errorf("phase1 fallback: %w", err)
+			}
+			maps.Copy(accumulated, out)
+		}
+	}
+
+	// Build implMap: for each blockchain deployed in Phase 1, instantiate its
+	// empty CCIP17Configuration using the registered NetworkConfigFactory.
```
winder, line 65: MK: If you can build the impl in a phase that would be nice.

build/devenv/components/blockchains/component.go
```diff
@@ -0,0 +1,109 @@
...
+	if err := devenvruntime.Register(configKey, factory); err != nil {
+		panic(fmt.Sprintf("blockchains component: %v", err))
+	}
+}
+
+func factory(_ map[string]any) (devenvruntime.Component, error) {
+	return &component{}, nil
+}
+
+type component struct{}
```
winder, line 34: This component could have an optional `GetImpls()` function that allows the runtime to get constructed CCIP17Configuration objects

build/devenv/cciptestinterfaces/interface.go
winder, line 368: All of these can go away. They are replaced by specific components. Then we don't have to keep track of "Impl" objects.

build/devenv/runtime/component.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Phase3Component runs during Phase 3 (CCVs and token pools).
+type Phase3Component interface {
+	RunPhase3(ctx context.Context, globalConfig map[string]any, componentConfig any, priorOutputs map[string]any, implMap map[string]cciptestinterfaces.CCIP17Configuration) (map[string]any, error)
+}
+
+// Phase4Component runs during Phase 4 (final configuration).
+type Phase4Component interface {
+	RunPhase4(ctx context.Context, globalConfig map[string]any, componentConfig any, priorOutputs map[string]any, implMap map[string]cciptestinterfaces.CCIP17Configuration) error
+}
```
winder, line 61: These phases are codifying the DAG. But what if you need something in between 3 and 4?

The DAG could be built with an API that knows blockchain comes first.
* Unix often will use sort order and multiple configs "10blockchain", "20jobdistributor", ...

Pro's: this approach is simple and easy to adopt.
Con's: it is not flexible enough for every use case, there could be times when the order of components in 1 phase is important (i.e. it must run between 3 and 4).

build/devenv/runtime/environment.go
winder: After further discussion, the ImplMap can go away entirely, the functions would be migrated to the components similar to how the Deploy function was migrated.

build/devenv/cli/ccv.go
```diff
@@ -49,6 +52,18 @@ var rootCmd = &cobra.Command{
...
 		}
+		mode, err := cmd.Flags().GetString("env-mode")
+		if err != nil {
+			return err
+		}
+		switch mode {
+		case "legacy":
+			newEnvFn = ccv.NewEnvironment
+		case "phased":
+			newEnvFn = ccv.NewEnvironmentPhased
```
winder, line 63: The new phased environment is a fork of the original environment. This is good for a gradual migration, but puts us at risk of having the environments diverge.

PR #1076, opened 2026-05-05
Disable development flag for logging

carte7000:
## Checklist


makramkd (dismissed):

asoliman92 (dismissed):

winder (changes requested):
Isn't the real fix to stop using `DevelopmentConfig` in production? Don't we still want this behavior for devenv and other test deployments?

makramkd:
@winder technically yes, though I think we can do that in a follow-up.

winder:
@makramkd we should make the proper change. In the worst case this could prevent unit tests from failing and allow regressions to slip in.

winder (dismissed):

winder (approved):

asoliman92 (approved):

PR #1077, opened 2026-05-05
Add ListMessageRules client

carte7000:
## Checklist

### DIFF
--- a/common/messagerules/proto.go
+++ b/common/messagerules/proto.go
@@ -1,0 +1,151 @@
+package messagerules
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-ccv/protocol"
+	messagepb "github.com/smartcontractkit/chainlink-protos/chainlink-ccv/message-rules/v1"
+)
+
+func RuleToProto(rule Rule) (*messagepb.MessageRule, error) {
+	pbRule := &messagepb.MessageRule{
+		Id:                  rule.ID,
+		CreatedAtUnixMillis: timeToUnixMillis(rule.CreatedAt),
+		UpdatedAtUnixMillis: timeToUnixMillis(rule.UpdatedAt),
+	}
+
+	switch rule.Type {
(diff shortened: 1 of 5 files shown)


common/messagerules/proto.go
```diff
@@ -0,0 +1,151 @@
...
+	return pbRule, nil
+}
+
+func unixMillis(millis int64) time.Time {
+	if millis == 0 {
+		return time.Time{}
+	}
+	return time.UnixMilli(millis).UTC()
+}
+
```
asoliman92, line 72: we have these time conversion functions duplicated in a couple of protos.go . should we have a common one and resuse them better?

common/messagerules/rules.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/smartcontractkit/chainlink-ccv/protocol"
+)
+
+type CompiledRules struct {
+	rules         []activeRule
+	rulesSnapshot []Rule
+	counts        RuleCounts
+}
+
+type RuleCounts struct {
```
asoliman92: what do we use these for exactly? does the number of each type matter or give us needed info?

carte7000: I am planning on using this for metrics later on in aggregator

common/messagerules/proto.go
carte7000: I think the only other place we have this helper is in the agg proto. I will move it to this package in another PR but that will be a bigger change cause I need to make aggregator reuse the rules from the integration package as well

common/messagerules/rules.go
asoliman92: I'm not sure if it's a really useful metric if we don't know details like which chains it's on or something like that. Just counts is not very helpful for debugging. Not a blocker though but I think it might bloat the code unnecessarily.

asoliman92 (dismissed):

asoliman92 (approved):

crispymangoes (approved):

PR #1084, opened 2026-05-06
Add token receiver to generic extra arg v3

huangzhen1997:
Support token receiver for the extra args V3 type

### DIFF
--- a/build/devenv/cciptestinterfaces/extra_args.go
+++ b/build/devenv/cciptestinterfaces/extra_args.go
@@ -20,6 +20,8 @@ type MessageOptions struct {
 	Executor protocol.UnknownAddress
 	// ExecutorArgs are the executor arguments for the message
 	ExecutorArgs []byte
+	// TokenReceiver is the token receiver address, used when token receiver different from the message receiver
+	TokenReceiver []byte
 	// TokenArgs are the token arguments for the message
 	TokenArgs []byte
 }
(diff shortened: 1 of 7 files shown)


makramkd (approved):

build/devenv/cciptestinterfaces/extra_args.go
```diff
@@ -47,12 +55,13 @@ type MessageV3Destination interface {
...
 // We use an interface rather than a struct because the V3 message structure is chain agnostic.
 type MessageV3Source interface {
-	// BuildV3ExtraArgs builds the V3 extra arguments for the message including calling the destination chain's GetExecutorArgs and GetTokenArgs.
+	// BuildV3ExtraArgs builds the V3 extra arguments for the message including calling the destination chain's GetExecutorArgs, GetTokenReceiver and GetTokenArgs.
 	// then serializing the results into the source chain specific encoding format, and returning the result.
 	BuildV3ExtraArgs(
 		opts MessageOptions,
 		destChain MessageV3Destination,
 		executorArgsParams any,
+		tokenReceiverParams any,
```
winder, line 64: This field is already in the `MessageOptions` object, why add it as another field? My understanding is that this pattern is only needed for non-standard fields like "extra args".

huangzhen1997: The MessageOptions is the extraArgs itself ([GenericExtraArgsV3](https://github.com/smartcontractkit/chainlink-ccv/blob/0eae0cc53a9a18697c23a6924007c95eddf1d802/build/devenv/cciptestinterfaces/extra_args.go#L8)). Base on the existing implementation, the MessageOptions.ExecutorArgs and executorArgsParams have different purpose. 

The MessageOptions.[ExecutorArgs](https://github.com/smartcontractkit/chainlink-ccv/blob/0eae0cc53a9a18697c23a6924007c95eddf1d802/build/devenv/evm/impl.go#L798) is the eventual raw bytes form before sourceChain encoding, and the [executorArgsParams](https://github.com/smartcontractkit/chainlink-ccv/blob/0eae0cc53a9a18697c23a6924007c95eddf1d802/build/devenv/evm/impl.go#L2089) is the input for GetExecutorArgs to perform destination chain encoding, and then pass to the MessageOptions

build/devenv/cciptestinterfaces/extra_args.go
ogtownsend, line 9: nit: just say explicitly `MessageOptions is GenericExtraArgsV3` until it's renamed, "extra args" is so overloaded now

```suggestion
// MessageOptions is the same thing as GenericExtraArgsV3 and represents modifications one can make to a CCIP messagev3 through extra args.
// TODO: rename this to GenericExtraArgsV3.
type MessageOptions struct {
```

build/devenv/cciptestinterfaces/extra_args.go
ogtownsend: Feel like we should take a page out of java's book and go crazy verbose with naming variables...at least when it comes to `ExtraArgs`... It'd be nice if all devenv-specific types were prefixed with `Devenv` to help distinguish them from their actual protocol counterparts. :|

ogtownsend (approved):

winder (approved):

PR #1089, opened 2026-05-07
Introduce phased devenv runtime alongside the monolith

winder:
## Description
First incremental slice of the phased devenv design (design.txt). Lands the runtime scaffolding without extracting any concrete component yet:

 - build/devenv/runtime/ — Component + Phase 1–4 interfaces, Registry with fallback support, and the phase driver. Phase 4 returns map[string]any (merged into the accumulator) so future post-Phase-4 consumers have access to its outputs. No implMap / RegisterNetworkFactory plumbing in this PR.
 - environment_monolith.go — existing NewEnvironment moved verbatim. Stable reference path; will not change as components are extracted.
 - environment_phased.go — NewPhasedEnvironment entry point loads raw TOML and drives the runtime. runPhasedEnvironment is a fork of the monolith body that the fallback embeds; it shrinks as components are extracted in follow-ups.
 - legacy_component.go — fallback component registered via runtime.SetFallback; its RunPhase2 invokes runPhasedEnvironment and exposes the resulting *Cfg to NewPhasedEnvironment.
 - config.go — adds loadRaw (TOML → map[string]any) for the runtime.
 - cli/ccv.go — --env-mode {legacy|phased} flag (default legacy) selects between NewEnvironment and NewPhasedEnvironment for ccv up / ccv restart.
 - .github/workflows/test-smoke.yaml — new TestE2ESmoke_Basic_Phased matrix entry that brings the env up with --env-mode phased and reuses the
 TestE2ESmoke_Basic test.

 Legacy mode is unchanged. Phased mode currently runs the entire monolith via the fallback; subsequent PRs will carve out blockchain, indexer, executor, etc. as concrete components.

## Checklist


build/devenv/runtime/environment.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	if fallback != nil {
+		if p2, ok := fallback.(Phase2Component); ok {
+			out, err := p2.RunPhase2(ctx, rawConfig, unclaimed, maps.Clone(accumulated))
+			if err != nil {
+				return nil, fmt.Errorf("phase2 fallback: %w", err)
+			}
+			maps.Copy(accumulated, out)
+		}
+	}
```
makramkd, line 80: **nit**: I wonder if this loop + if statement can be abstracted to a function that just gets called 4 times with different methods. Though I guess that might use reflection, which could make things less readable.

makramkd (approved):

build/devenv/runtime/environment.go
winder: Totally agree. I'm expecting this entire function to be rewritten multiple times. I'd also like to collect the function pointers into 4x `[]Phase{1,2,3,4}Component` objects at the beginning to simplify the functions even further.

One nice thing about hiding all the business logic inside components is that these sorts of algorithms become very easy to refactor.

carte7000 (approved):

PR #1090, opened 2026-05-08
devenv: snapshot accumulated outputs per phase and reject overwrites.

winder:
## Description

Stricter control and validation around inputs and outputs to the devenv phased runtime.
* Snapshots are taken at the start of each phase to avoid polluting the accumulated outputs across components.
* Check for value overrides when capturing outputs.

## Testing
New unit tests.

## Checklist


build/devenv/.mockery.yaml
makramkd, line 1: Does this need to be checked in CI w/ its own action?

winder: Nope, CI uses find and runs mockery on every directory with a .mockery.yml file.

makramkd (approved):

build/devenv/runtime/environment.go
```diff
@@ -37,93 +50,135 @@ func NewEnvironmentWithRegistry(ctx context.Context, rawConfig map[string]any, r
...
+// already exists in dst. The phase number and owner identify the offending
+// component in the error message.
+func mergeNoOverwrite(dst, src map[string]any, phase int, owner string) error {
+	for k, v := range src {
+		if _, exists := dst[k]; exists {
+			return fmt.Errorf(
+				"phase %d component %q wrote output key %q that is already set; "+
+					"same-phase components must not collide and components must not "+
+					"overwrite outputs from prior phases",
+				phase, owner, k)
```
jadepark-dev, line 175: nit: would it be useful to also include which component/phase originally set the key for easier debugging?

jadepark-dev (approved):

build/devenv/runtime/environment.go
winder: Maybe I can add a type to the key to make it clearer that it's also a unique ID for the component. In theory this case should never trigger.

I'm mainly concerned about the "legacy" component while I'm migrating to the phased environment. It's much more ad-hoc, so I wanted to add this check just in case I miss something.

PR #1092, opened 2026-05-09
tcapi: simplified interface

makramkd:
## Description

The `tcapi.TestCase` interface's methods were depending on `ccv.Cfg` and `TestHarness`, both of which make it difficult to run test cases on non-ephemeral environments like CLD environments.

This is the first step in making the test cases environment-agnostic:

* `tcapi.TestCase` methods now only take a `context.Context`, typically the test context. 
* The _construction_ of test cases will take a `deployment.Environment` and functional options that specify things like aggregator and indexer clients. This is because in CLD such clients are constructed from URLs that are retrieved from outside the datastore, compared to the URLs being in the `ccv.Cfg` in ephemeral envs.
* The `basic` test cases now make the aggregator and indexer guards optional; i.e. if aggregator and indexer clients are made available via the requisite functional options, the guards are enabled, else they are disabled. This kind of behaviour should probably be applied to all future `TestCase` implementations.

The next steps will be:
1. [P0] Migrate tcapi basic + token_transfer send path off SendMessage to BuildChainMessage + SendChainMessage. Aligns with the deprecation on Chain; same behavior, future-proof surface.
2. [P1] Refactor remaining e2e helpers that still center Lib + Cfg where tcapi-style env wiring is preferred, e.g. token verification / extra-args v2 tests: same explicit lane + NewCLDFOperationsEnvironment + optional clients pattern as smoke, where it reduces duplication without rewriting full v2 chaos/load.
3. [P1] Evaluate an optional TestEnvironment façade (DataStore + CLDFEnvironment). Single adapter + NewTestEnvironmentFromCfg would narrow what tcapi (and tests) depend on if CLDF churn is a concern.
4. [P2] Reconcile composable scenarios (MessageV3TestScenario, v2/v3 helpers) with tcapi.
5. [P2] After (4): revisit typing src / dst as ChainAsSource / ChainAsDestination (or a composed test-facing interface)

## Checklist


build/devenv/tests/e2e/tcapi/options.go
```diff
@@ -0,0 +1,119 @@
...
+
+// CaseDeps holds dependencies wired at construction time for TestCase implementations.
+type CaseDeps struct {
+	Env               *deployment.Environment
+	DataStore         datastore.DataStore
+	SrcSelector       uint64
+	DstSelector       uint64
+	AggregatorClients map[string]*ccv.AggregatorClient
+	IndexerMonitor    *ccv.IndexerMonitor
+	ChainMap          map[uint64]cciptestinterfaces.CCIP17
```
makramkd, line 27: The use of the very wide `CCIP17` interface will be addressed in follow-ups.

build/devenv/tests/e2e/tcapi/options.go
```diff
@@ -0,0 +1,119 @@
...
+type CaseOption func(*caseConfig) error
+
+// WithLane selects the source and destination chain selectors for lane-scoped tests.
+func WithLane(srcSel, dstSel uint64) CaseOption {
+	return func(c *caseConfig) error {
+		c.srcSel, c.dstSel = srcSel, dstSel
+		c.laneSet = true
+		return nil
+	}
+}
```
makramkd, line 50: Right now WithLane is an option but is actually required. I think we can fall back to a random arrangement of (src, dest) for whatever selectors are available in the deployment.Environment if its not provided.

build/devenv/tests/e2e/tcapi/options.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	lg := zerolog.Ctx(ctx)
+	src, err := ccv.NewCCIP17ForChainSelector(ctx, *lg, env, rc.srcSel)
+	if err != nil {
+		return nil, fmt.Errorf("src chain selector %d: %w", rc.srcSel, err)
+	}
+	dst, err := ccv.NewCCIP17ForChainSelector(ctx, *lg, env, rc.dstSel)
+	if err != nil {
+		return nil, fmt.Errorf("dst chain selector %d: %w", rc.dstSel, err)
+	}
```
makramkd, line 105: This constructor is now decoupled from ccv.Cfg which is good, nothing to change here.

build/devenv/chain_impl_factory.go
```diff
@@ -21,14 +20,13 @@ type ImplFactory interface {
...
-	// tests.
+	// New creates a new cciptestinterfaces.CCIP17 object from the deployment environment
+	// and chain selector (must exist on env.BlockChains for the implementation's family).
 	New(
 		ctx context.Context,
-		cfg *Cfg,
 		lggr zerolog.Logger,
 		env *deployment.Environment,
-		bc *blockchain.Input,
+		chainSelector uint64,
```
makramkd, line 29: Breaking change here to only rely on deployment.Environment which should have all blockchain clients anyway.

build/devenv/ccip17_factory.go
```diff
@@ -0,0 +1,27 @@
...
+
+	chain_selectors "github.com/smartcontractkit/chain-selectors"
+
+	"github.com/smartcontractkit/chainlink-ccv/build/devenv/cciptestinterfaces"
+	"github.com/smartcontractkit/chainlink-deployments-framework/deployment"
+)
+
+// NewCCIP17ForChainSelector builds a CCIP17 implementation for the given selector using the
+// registered ImplFactory for that selector's chain family.
+func NewCCIP17ForChainSelector(ctx context.Context, lggr zerolog.Logger, env *deployment.Environment, chainSelector uint64) (cciptestinterfaces.CCIP17, error) {
```
winder, line 17: **comment**: A lot of these helpers seem like they're pretty much the same as the behavior as the stuff in `lib.go`. We should try and consolidate that -- either with an interface with separate `env-out.toml` and `deployment.Environment` constructors, or by picking one approach.

**question**: Is that what you had in mind with these followup tasks?
```
[P1] Refactor remaining e2e helpers that still center Lib + Cfg where tcapi-style env wiring is preferred, e.g. token verification / extra-args v2 tests: same explicit lane + NewCLDFOperationsEnvironment + optional clients pattern as smoke, where it reduces duplication without rewriting full v2 chaos/load.
[P1] Evaluate an optional TestEnvironment façade (DataStore + CLDFEnvironment). Single adapter + NewTestEnvironmentFromCfg would narrow what tcapi (and tests) depend on if CLDF churn is a concern.
```

build/devenv/chain_impl_factory.go
winder: I'm also changing this function. [My PR](https://github.com/smartcontractkit/chainlink-ccv/pull/1108) is also removing `cfg` and moved it to a new package.

PR #1096, opened 2026-05-11
[Chore] refactor devenv (transmitter) key funding

faisal-chainlink:
## Description

This issue this PR solves is that devenv previously ignored non-EVM keys when bringing up a bootstrapped service. The `#FetchBootstrapKeys` method tries to handle conversion of public keys to addresses which works well for EVM but would require hardcoding some keys names in `chainlink-ccv` if we want to apply the same approach for altVMs (specifically ones that use a different address format than than of EVM).

The changes are primarily concerned without allowing each impl factory to return the name of the default key (name) used for transmission as well as exposing a method for deducing the address from a `keystore` key response.

Finally, the EVM factory and other EVM impls are updated to handle the change described above.

## Testing

1. E2E tests passing
2. Corresponding PR in `chainlink-stellar` uses the changes defined in this PR

## Checklist

...TODO


build/devenv/chain_impl_factory.go
```diff
@@ -37,6 +38,18 @@ type ImplFactory interface {
...
+	// transaction transmission on this chain family (e.g.,
+	// executor.DefaultEVMTransmitterKeyName for EVM,
+	// common.StellarTransmitterKeyName for Stellar).
+	DefaultTransmitterKeyName() string
+
+	// DeriveAddressesFromKeys extracts protocol addresses from bootstrap keys
+	// for this chain family. It should look up its DefaultTransmitterKeyName
+	// in keys.Keys and convert the public key to the appropriate address format.
+	// Returns nil if the key is not found.
+	DeriveAddressesFromKeys(keys services.BootstrapKeys) []protocol.UnknownAddress
```
faisal-chainlink, line 51: perhaps it's better to place this within the ChainFamilyAdapter rather than here?

winder: We should definitely avoid adding functions here. Ideally it goes away completely. Functions can go in `CCIP17` or as a functional adapter.

build/devenv/services/bootstrap.go
```diff
@@ -101,6 +101,9 @@ type BootstrapKeys struct {
 	// EVMTransmitterAddress is the Ethereum address derived from the executor's EVM transmitter key.
 	// Only populated for executor nodes.
 	EVMTransmitterAddress string `toml:"evm_transmitter_address,omitempty"`
+	// Keys contains ALL fetched keystore responses keyed by name.
+	// This allows chain families to access their specific key types generically.
+	Keys map[string]keystore.GetKeyResponse `toml:"keys,omitempty"`
```
winder, line 106: **nit**: I'm generally hesitant to add a dependency just for a type.

**question**: Should this be `map[string][]keystore.GetKeyResponse`?

**question**: Would the goal be to drop the CSA/ECDSA/EVM fields eventually?

PR #1106, opened 2026-05-13
CCIP-11383: Improved accessor lifecycle support

jadepark-dev:
## Description

Adds `Close() error` to `chainaccess.Accessor` and `bootstrap.AccessorCloserRegistry` to track and close all accessors at shutdown. Unblocks Solana log poller cleanup (NONEVM-5006, smartcontractkit/chainlink-ccip-solana#135).

- `AccessorCloserRegistry` wraps Registry, tracks accessors via `GetAccessor`, closes all via `CloseAll`
- Wired in both JD and static-config shutdown paths; deferred cleanup on partial-Start failure
- EVM accessor `Close()` is a no-op
- **Breaking:** external `Accessor` implementers must add `Close() error`

## Testing

- `go test ./bootstrap/... -v` — 6 accessor lifecycle tests pass (close-all, error aggregation, concurrency, ordering, partial-start cleanup, static-config stop)
- `go build ./bootstrap/...` — clean
- `just mock` — regenerated

## Checklist


bootstrap/accessor_tracker.go
```diff
@@ -0,0 +1,71 @@
...
+//
+// Callers must invoke CloseAll after factory.Stop (or its equivalent) returns,
+// so the factory's coordinator drains its readers before the underlying log
+// pollers are torn down.
+type AccessorTracker struct {
+	lggr  logger.Logger
+	inner chainaccess.Registry
+
+	mu        sync.Mutex
+	accessors []chainaccess.Accessor
```
makramkd: **question**: wrapper object vs. having this tracking logic in the registry itself; did you think about that and what the pros and cons are for each?

bootstrap/accessor_tracker.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// NewAccessorTracker wraps inner so every successful GetAccessor result is tracked.
+func NewAccessorTracker(lggr logger.Logger, inner chainaccess.Registry) *AccessorTracker {
+	return &AccessorTracker{lggr: lggr, inner: inner}
+}
+
+// GetAccessor delegates to the inner Registry and tracks the returned Accessor.
+func (t *AccessorTracker) GetAccessor(ctx context.Context, chainSelector protocol.ChainSelector) (chainaccess.Accessor, error) {
+	a, err := t.inner.GetAccessor(ctx, chainSelector)
```
makramkd: **question**: not sure the likelihood of this happening, but I believe the way the code is currently constructed now, if you call `GetAccessor` after the critical section in `CloseAll` happens, you'd end up eventually adding an accessor to `t.accessors` while the `CloseAll` is happening, i.e. you might miss an accessor to close. I'm wondering if there's a way to mitigate this?

I believe if you make the critical section the entire method you won't be able to interleave:

```golang
func (t *AccessorTracker) GetAccessor(ctx context.Context, chainSelector protocol.ChainSelector) (chainaccess.Accessor, error) {
  t.mu.Lock()
  defer t.mu.Unlock()
  
  // same current logic, create accessor, check err, return, add to slice
}

func (t *AccessorTracker) CloseAll() error {
  t.mu.Lock()
  defer t.mu.Unlock()

  // same current logic, close, collect errors, return
}
```

I believe this way you won't run into the same issue because GetAccessor and CloseAll can't interleave at all.

bootstrap/accessor_tracker.go
winder: The pro is clear:
* Resources are always cleaned up.

The only cons I see are fairly speculative:
* The service factory code is not a good reference for how to use the registry, if someone looks at this code while initializing an accessor for a test suite they may not know about the resource cleanup requirements.
* If the bootstrap evolves to support multiple concurrent apps, we'd have to make sure there is a separate accessor tracker per app to avoid lifecycle collisions.

bootstrap/accessor_tracker.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+)
+
+// AccessorTracker wraps a chainaccess.Registry, tracks every Accessor handed
+// out, and closes them all via CloseAll at shutdown.
+//
+// Callers must invoke CloseAll after factory.Stop (or its equivalent) returns,
+// so the factory's coordinator drains its readers before the underlying log
+// pollers are torn down.
+type AccessorTracker struct {
```
winder: **nit**: Maybe **AccessorCloserRegistry**? I think it's good to name things by why you'd want to use it rather than how it implements the functionality.

bootstrap/accessor_tracker.go
jadepark-dev: Yeah, putting tracking in the registry itself felt like mixing resource lifecycle with construction.

bootstrap/accessor_tracker.go
jadepark-dev: Sounds good. I'll make it more defensive

bootstrap/accessor_tracker.go
jadepark-dev: Much better, thanks!

makramkd (approved):

bootstrap/accessor_closer_registry.go
```diff
@@ -0,0 +1,71 @@
...
+		if err := a.Close(); err != nil {
+			errs = append(errs, fmt.Errorf("close accessor[%d]: %w", i, err))
+		}
+	}
+
+	if err := errors.Join(errs...); err != nil {
+		t.lggr.Warnw("some accessors failed to close", "error", err)
+		return err
+	}
+
```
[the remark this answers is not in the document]
jadepark-dev: all four call sites are terminal shutdown paths with no retry loop

pkg/chainaccess/interfaces.go
```diff
@@ -82,6 +82,8 @@ type Accessor interface {
 	DestinationReader() (DestinationReader, error)
 	// ContractTransmitter returns the ContractTransmitter for this chain, or an error if not available.
 	ContractTransmitter() (ContractTransmitter, error)
+	// Close releases any background services the accessor started. Stateless accessors return nil.
```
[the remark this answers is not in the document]
jadepark-dev: every factory create a fresh accessor per GetAccessor call

winder (approved):

bootstrap/accessor_tracker.go
ogtownsend: oh I think this is related to the question I asked here right? https://github.com/smartcontractkit/chainlink-ccip-solana/pull/135#discussion_r3245934081

PR #1109, opened 2026-05-13
add ccv specific deploy

tt-cll:
## Checklist

### DIFF
--- a/deployment/adapters/committee_verifier_deploy.go
+++ b/deployment/adapters/committee_verifier_deploy.go
@@ -29,28 +29,23 @@ type CommitteeVerifierDeployParams struct {
 // responsible for interpreting DeployerContract (e.g. a CREATE2Factory hex
 // address on EVM) and for looking up family-specific dependencies (such as
 // the RMNProxy) from ExistingAddresses.
+//
+// Deployed contracts remain owned by the deployer key. Transferring
+// ownership to a timelock/MCMS is composed by a separate changeset on top
+// (see CCIP-11432).
 type DeployCommitteeVerifierInput struct {
 	ChainSelector     uint64
 	DeployerContract  string
 	ExistingAddresses []datastore.AddressRef
 	Params            CommitteeVerifierDeployParams
-	// DeployerKeyOwned, when true, leaves the deployed contracts owned by the
-	// deployer key. When false, the adapter is expected to surface the
-	// contracts that need ownership transfer in RefsToTransferOwnership so
-	// the caller can wire up MCMS-based transfer. The current chain-agnostic
-	// changeset only supports DeployerKeyOwned=true.
-	DeployerKeyOwned bool
 }
 
 // DeployCommitteeVerifierOutput is the chain-agnostic output of a
-// CommitteeVerifier deploy. Adapters return any newly deployed addresses,
-// MCMS batch operations produced by post-deploy configuration writes, and
-// the subset of contracts that still need ownership transfer when not
-// running in deployer-key-owned mode.
+// CommitteeVerifier deploy. Adapters return any newly deployed addresses
+// and MCMS batch operations produced by post-deploy configuration writes.
 type DeployCommitteeVerifierOutput struct {
-	Addresses               []datastore.AddressRef
-	BatchOps                []mcmstypes.BatchOperation
-	RefsToTransferOwnership []datastore.AddressRef
+	Addresses []datastore.AddressRef
+	BatchOps  []mcmstypes.BatchOperation
 }
 
 // CommitteeVerifierDeployAdapter is the chain-family-specific entry point
--- a/deployment/adapters/committee_verifier_deploy.go
+++ b/deployment/adapters/committee_verifier_deploy.go
@@ -2,11 +2,12 @@ package adapters
 
 import (
 	"github.com/Masterminds/semver/v3"
-	mcms_types "github.com/smartcontractkit/mcms/types"
 
-	cldf_chain "github.com/smartcontractkit/chainlink-deployments-framework/chain"
+	mcmstypes "github.com/smartcontractkit/mcms/types"
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain"
 	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
-	cldf_ops "github.com/smartcontractkit/chainlink-deployments-framework/operations"
+	"github.com/smartcontractkit/chainlink-deployments-framework/operations"
 )
 
 // CommitteeVerifierDeployParams describes the per-committee deployment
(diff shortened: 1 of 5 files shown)


deployment/changesets/deploy_committee_verifier.go
```diff
@@ -0,0 +1,193 @@
...
+			if !slices.Contains(envSelectors, sel) {
+				return fmt.Errorf("chain selector %d is not available in environment", sel)
+			}
+
+			perChain := cfg.resolveChainCfg(sel)
+			if perChain.DeployerContract == "" {
+				return fmt.Errorf("DeployerContract is required for chain %d", sel)
+			}
+
+			a, err := registry.GetByChain(sel)
```
RensR: Please do not use single char names

deployment/changesets/deploy_committee_verifier.go
```diff
@@ -0,0 +1,193 @@
...
+		for _, sel := range cfg.ChainSelectors {
+			if seen[sel] {
+				return fmt.Errorf("duplicate chain selector %d in ChainSelectors", sel)
+			}
+			seen[sel] = true
+			if !slices.Contains(envSelectors, sel) {
+				return fmt.Errorf("chain selector %d is not available in environment", sel)
+			}
+
+			perChain := cfg.resolveChainCfg(sel)
```
RensR: nit: perChain name isn't super descriptive but maybe we can just inline it since it's only used once?

```
			if cfg.resolveChainCfg(sel).DeployerContract == "" {
				return fmt.Errorf("DeployerContract is required for chain %d", sel)
			}
```

deployment/changesets/deploy_committee_verifier.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		envSelectors := e.BlockChains.ListChainSelectors()
+		seen := make(map[uint64]bool, len(cfg.ChainSelectors))
+		for _, sel := range cfg.ChainSelectors {
+			if seen[sel] {
+				return fmt.Errorf("duplicate chain selector %d in ChainSelectors", sel)
+			}
+			seen[sel] = true
+			if !slices.Contains(envSelectors, sel) {
```
RensR, line 83: I'd expect checks like these to be very generic and used in a lot of places. Have we considered extracting this into a single package?

deployment/adapters/committee_verifier_deploy.go
```diff
@@ -0,0 +1,72 @@
...
+	ChainSelector     uint64
+	DeployerContract  string
+	ExistingAddresses []datastore.AddressRef
+	Params            CommitteeVerifierDeployParams
+	// DeployerKeyOwned, when true, leaves the deployed contracts owned by the
+	// deployer key. When false, the adapter is expected to surface the
+	// contracts that need ownership transfer in RefsToTransferOwnership so
+	// the caller can wire up MCMS-based transfer. The current chain-agnostic
+	// changeset only supports DeployerKeyOwned=true.
+	DeployerKeyOwned bool
```
RensR: If we only support one value do we really need this? Do we foresee many cases where we want to have this contract be owned by someone else and cannot compose a transferOwnership on top of it?

deployment/changesets/deploy_committee_verifier_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+}
+
+func TestDeployCommitteeVerifier_Validation_NoChainSelectors(t *testing.T) {
+	cs := DeployCommitteeVerifier(newDeployTestRegistry())
+	err := cs.VerifyPreconditions(newDeployTestEnv(nil), DeployCommitteeVerifierInput{
+		Committees: []adapters.CommitteeVerifierDeployParams{validCommittee()},
+	})
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "at least one chain selector is required")
```
RensR: Prefer 

```
	require.ErrorContains(t, err, "at least one chain selector is required")
```

More instances below

deployment/changesets/deploy_committee_verifier.go
```diff
@@ -0,0 +1,193 @@
...
+
+// DeployCommitteeVerifierInput is the imperative input for the chain-agnostic
+// DeployCommitteeVerifier changeset. Callers describe the committees they
+// want deployed plus the chains those committees should be deployed onto;
+// the per-chain adapter resolves any chain-family-specific dependencies
+// (RMN, factories, etc.) out of the environment DataStore.
+//
+// One CommitteeVerifier instance is deployed per (chain, committee) pair —
+// i.e. len(ChainSelectors) × len(Committees) deploys in total.
+type DeployCommitteeVerifierInput struct {
```
RensR, line 31: Why do we need 

- adapters.CommitteeVerifierDeployParams
- adapters.DeployCommitteeVerifierInput
- changesets.DeployCommitteeVerifierPerChainCfg
- changesets.DeployCommitteeVerifierInput

Feels like a lot, and especially the two DeployCommitteeVerifierInput types are confusing

deployment/changesets/deploy_committee_verifier.go
tt-cll: I think it is something we should end up doing. Historically we haven't had these checks in many changesets but I think it's a good hardening technique to push errors up the stack rather than runtime. https://smartcontract-it.atlassian.net/browse/CCIP-11468

deployment/adapters/committee_verifier_deploy.go
tt-cll: This was mostly just done for parity with the existing `DeployContracts`, but I agree with your intuition. Ideally the deploy + ownership is set as an all-in-one to avoid misconfiguration in prod. For protocol contracts, it's pretty linear as it will always be timelock as the owner. 

I don't know what the operational plan is for customer CCVs. Theoretically, they can (should?) deploy and own them, but I do wonder if growth ends up deploying and managing some subset of them, similar to token pools.

We can't wire up transfers here anyway yet e2e (CLD is refactoring and we will follow up here afterwards https://smartcontract-it.atlassian.net/browse/CCIP-11432) so we can remove the param for now and layer transfer when ready. The use case this changeset is unblocking for now is just in-memory devenv optimization, so deployer only is fine.

deployment/changesets/deploy_committee_verifier.go
tt-cll: Yeah I agree. This wasn't intentional to have the split and was mostly just conforming to the existing pattern (which was carried over from the original 2.0 changesets). I will look to consolidate in a followup https://smartcontract-it.atlassian.net/browse/CCIP-11469

RensR (approved):
Approved given the comments stating some follow up improvements

winder (approved):

PR #1117, opened 2026-05-18
build/devenv: consolidate disparate registries

makramkd:
## Description

Devenv is currently littered with many different registries which makes discovery of them difficult and introduces potentially different registry APIs within the same Go module.

In this PR, we:

* Consolidate devenv chain-family registration into the new build/devenv/chainreg package.
* Replace separate registries for impl factories, CLDF providers, chain config loaders, launchers, modifiers, and extra-args serializers.
* Move EVM registration pieces into build/devenv/evm/registration.go.
* Remove global modifier registries and injects verifier/executor modifier maps into committeeverifier.New and executor.New.
* Move extra-args serializer storage/lookup from cciptestinterfaces into chainreg.
* Adds focused chainreg tests for serializer lookup and partial registration merging.

## Testing

Existing regression test suite in CI.

## Checklist


build/devenv/chainreg/types.go
```diff
@@ -0,0 +1,96 @@
...
+// Fields are optional; callers should set what the family supports.
+type Registration struct {
+	ImplFactory          ImplFactory
+	CLDFProvider         CLDFProviderFactory
+	ChainConfigLoader    ChainConfigLoader
+	Launcher             Launcher
+	VerifierModifier     VerifierModifier
+	ExecutorModifier     ExecutorModifier
+	ExtraArgsSerializers map[uint8]ExtraArgsSerializer
+}
```
winder, line 96: There's some benefit to loose coupling. Having this strongly typed integration that eases discovery of the implementations seems fine, but it makes this a difficult import.

Do you think it would be helpful to include a dynamic registration layer before the strongly typed layer?

i.e.
```
type Registry struct {
	mu            sync.RWMutex
	registrations map[protocol.ChainSelector]map[string]any
}
```

I guess that inner map would be something like:
```
registrations[EVM] = map[string]any {
    "ImplFactory": ImplFactory{},
    "CLDFProvider": CLDFProvider{},
   ...
}
```

winder: The flip-side of this is that the 1.6 plugin config is exactly like what you have -- but it's completely separate from the deployment registry

[InitializePluginConfig](https://github.com/smartcontractkit/chainlink/blob/1d3b91c36d2259d00ac4dc8bd093d51275ea584f/core/capabilities/ccip/ccipton/pluginconfig.go#L14)
```
// InitializePluginConfig returns a pluginConfig for TON chains.
func InitializePluginConfig(lggr logger.Logger, extraDataCodec ccipocr3.ExtraDataCodecBundle) ccipcommon.PluginConfig {
	return ccipcommon.PluginConfig{
		AddressCodec:       codec.NewAddressCodec(),
		CommitPluginCodec:  codec.NewCommitPluginCodecV1(),
		ExecutePluginCodec: codec.NewExecutePluginCodecV1(extraDataCodec),
		// TODO(EVM2TON): this is a temp fix for nil msgHasher access, should be using CCIPProvider msgHasher instead, https://smartcontract-it.atlassian.net/browse/NONEVM-2778
		MessageHasher:         codec.NewMessageHasherV1(logger.Sugared(lggr).Named(chainsel.FamilyTon).Named("MessageHasherV1"), extraDataCodec),
		ExtraDataCodec:        codec.NewExtraDataDecoder(),
		GasEstimateProvider:   ccipnoop.NewGasEstimateProvider(extraDataCodec), // TODO: implement
		TokenDataEncoder:      ccipnoop.NewTokenDataEncoder(),                  // TODO: implement
		CCIPProviderSupported: true,
	}
}
```

makramkd: I tried to colocate all the types into the same package so that the import of this package meant that there is no additional discovery needed.

Re: the `map[protocol.ChainSelector]map[string]any` suggestion, that would change the registry from a per-family to a per-selector registry.

makramkd: The other issue I have w/ the `any` there is that we're not sure what actual implementation was passed in, whereas lots of code in devenv expects a specific type. I think that means you'd have to start casting in a bunch of places.

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -311,40 +310,9 @@ type OnChainConfigurable interface {
...
-type ExtraArgsSerializerEntry struct {
-	Version uint8
-	Family  string
-}
-
-// RegisterExtraArgsSerializer registers an ExtraArgsSerializer for a chain family.
-// If the family is already registered, the call is a no-op to match the pattern
-// used by other registries in this repo (e.g. CLDFProviderRegistry, ImplFactory).
-// Product repos call this in their init() alongside other registrations.
-func RegisterExtraArgsSerializer(entry ExtraArgsSerializerEntry, serializer ExtraArgsSerializer) {
```
makramkd, line 331: Moved to be part of the new canonical registry.

build/devenv/chainreg/types.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/smartcontractkit/chainlink-ccv/build/devenv/util"
+	cldf_chain "github.com/smartcontractkit/chainlink-deployments-framework/chain"
+	"github.com/smartcontractkit/chainlink-deployments-framework/deployment"
+	ctfblockchain "github.com/smartcontractkit/chainlink-testing-framework/framework/components/blockchain"
+)
+
+// ImplFactory is a factory for creating CCIP17 implementations.
+// Product repos (chainlink-canton, chainlink-stellar, chainlink-ccip-solana)
+// implement this interface and register it via Registration.ImplFactory in their init().
+type ImplFactory interface {
```
makramkd, line 22: Colocating as much types as I can into this package so that we reduce import cycle risk.

build/devenv/evm/impl.go
```diff
@@ -93,45 +90,7 @@ const (
...
-
-	tokenPoolVersions = []string{
-		"1.6.1",
-		"2.0.0",
-	}
-)
-
-// init registers evm token adapters for pool versions 1.6.1 and 2.0.0 so that ConfigureTokensForTransfers
-// (called from environment.go) can process token configs that reference these pool versions.
-func init() {
```
makramkd, line 107: Moved init to registration.go

tt-cll (approved):

winder (approved):
Looks like an incremental improvement to me. Nice 👍

PR #1120, opened 2026-05-18
build/devenv: extract contract deploy component

winder:
## Description

* Commit 1: build/devenv: move JD component from Phase 2 to Phase 1
  JD has no Phase 1 dependencies — it only starts a container and creates a client. Moving it to Phase 1
  unblocks the upcoming legacy Phase 2, which needs to read "jd" from priorOutputs to wire CL nodes and
  verifiers against JD before contract deployment runs in Phase 3.

## Checklist


build/devenv/components/monolith/aggregatorclient.go
```diff
@@ -0,0 +1,114 @@
...
+type AggregatorClient struct {
+	logger               zerolog.Logger
+	addr                 string
+	verifierResultClient verifierpb.VerifierClient
+	conn                 *grpc.ClientConn
+}
+
+// NewAggregatorClient creates a new AggregatorClient with TLS.
+// If caCertFile is provided, it will be used to verify the server certificate.
+func NewAggregatorClient(logger zerolog.Logger, addr, caCertFile string) (*AggregatorClient, error) {
```
makramkd: **question**: is this relevant to whether we spin up phased or monolith? I think this object is used mostly (exclusively?) in tests.

makramkd: I think this is a good type to have in a more top-level package that's not related to `monolith` - perhaps build/devenv/util or a new build/devenv/clients package.

build/devenv/components/monolith/cfg.go
```diff
@@ -0,0 +1,446 @@
...
+	// UseLegacyConfigureLane selects the legacy lanes.ConnectChains path
+	// instead of the canonical ConfigureChainsForLanesFromTopology changeset.
+	UseLegacyConfigureLane bool `toml:"use_legacy_configure_lane"`
+}
+
+type Cfg struct {
+	// Version is incremented on breaking config schema changes so downstream
+	// consumers can detect incompatible configs. Version 0 (implicit/absent)
+	// predates the [protocol_contracts] section.
+	Version            int                            `toml:"version"`
```
makramkd: **question**: This is a new field but its in the `monolith` package, which I imagine will be legacy. Is that intentional? Or will `monolith` eventually use the phased bringup?

build/devenv/components/protocol_contracts/component.go
```diff
@@ -0,0 +1,306 @@
...
+func (p *component) RunPhase3(
+	ctx context.Context,
+	_ map[string]any,
+	_ any,
+	priorOutputs map[string]any,
+) (map[string]any, []devenvruntime.Effect, error) {
+	configs := strings.Split(os.Getenv(ccv.EnvVarTestConfigs), ",")
+	if len(configs) > 1 {
+		ccv.L.Warn().Msg("Multiple configuration files detected, this feature may be unsupported in the future.")
+	}
```
makramkd, line 58: **question**: This is the input config(s) right? I wonder if this should be made as part of the RunPhase* method signatures to avoid this kind of logic (string split + reparse)?

build/devenv/components/protocol_contracts/component.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+		impl, ierr := ccv.NewProductConfigurationFromNetwork(bc.Type)
+		if ierr != nil {
+			return nil, nil, ierr
+		}
+		impls[i] = impl
+		blockchainOutputs[i] = bc.Out
+	}
+
+	in.CLDF.Init()
```
makramkd, line 126: **question**: how will CLDF/DataStore related transformations work across different components that need to mutate it? I guess we won't run the phases in the same level in parallel, so all updates should be serialized? But how do we ensure that there's no "clobbering" of previous updates, that sort of thing?

build/devenv/components/monolith/aggregatorclient.go
winder: Looks like it was dragged in because it was tied to `Cfg` via `NewAggregatorClientForCommittee`.

build/devenv/components/protocol_contracts/component.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	if len(in.Indexer) < 1 {
+		return nil, nil, fmt.Errorf("at least one indexer is required")
+	}
+
+	return map[string]any{
+		"aggregators":              in.Aggregator,
+		"_prepared_indexer_inputs": in.Indexer,
+		"_protocol_setup": &ccv.PhasedSetup{
```
makramkd, line 293: **question**: are the underscores here meant to signal that these are outputs that won't be part of the final output file?

build/devenv/environment_monolith.go
```diff
@@ -409,18 +409,6 @@ func NewEnvironment() (in *Cfg, err error) {
...
-	/////////////////////////////////////////
-
-	if err := launchGenericServices(ctx, in, e, blockchainOutputs); err != nil {
-		return nil, fmt.Errorf("failed to launch generic services: %w", err)
-	}
-
-	/////////////////////////////////////////
-	// END: Launch generic services //
-	/////////////////////////////////////////
-
```
makramkd, line 423: **question**: Will we put this back? It's currently used for Canton.

build/devenv/generic.go
makramkd, line 1: **question**: Same point here, will we revert this or was this moved somewhere?

build/devenv/components/protocol_contracts/component.go
winder: This is a separate AI context leaking into this session, so I can remove it.

Eventually I'd like slices of top-level configs to be handled by separate instances of the same component, so it's warning about that potential future state.

build/devenv/components/protocol_contracts/component.go
winder: Great point, I hadn't considered that. I think we need a `RegisterAddressEffect` that components can return to register new addresses with the runtime.

build/devenv/components/protocol_contracts/component.go
winder: It's currently being used to stage partial state updates across multiple components within the constraints I've put on the overall runtime. Specifically, the runtime doesn't allow overwriting values, so returning "indexer" here would cause a downstream error.

It should be cleaned up, but I left them in because they provide a nice mechanism to track dependencies across components without needing a strict dependency mechanism.

build/devenv/environment_monolith.go
winder: I thought it was unused, I'll put it back. In the future you'll add custom services by making a new component.

build/devenv/environment_phased.go
```diff
@@ -70,407 +66,11 @@ type phasedSetup struct {
...
 }
 
-// runPhasedEnvironmentSetup runs through aggregator config generation and
-// indexer config generation. Aggregator container launch is delegated to the
-// CommitteeCCV Phase 4 component, which reads "aggregators" and calls
-// services.NewAggregator, mutating the shared *AggregatorInput pointers.
-// Indexer container naming, TLS wiring, discovery config, and secrets are
-// delegated to the indexer Phase 4 component. runPhasedEnvironmentFinish
-// collects endpoints from the mutated Out fields on those shared pointers.
-func runPhasedEnvironmentSetup(ctx context.Context, in *Cfg) (*phasedSetup, error) {
```
winder, line 80: This function probably shouldn't have been created to begin with, it's gone entirely to the legacy component so that intermediate artifacts can be passed around via the phased runtime.

makramkd (approved):

tt-cll (approved):

PR #1121, opened 2026-05-19
tcapi: chain-agnostic address lookups

makramkd:
## Description

This PR modernizes the importable TCAPI test suite (basic, token_transfer) so it works across chain families without EVM-hardcoded contract lookups or deprecated SendMessage. It is the chainlink-ccv side of [chainlink-ccip-solana#190](https://github.com/smartcontractkit/chainlink-ccip-solana/pull/190) and unblocks downstream repos (Solana, Canton) that consume TCAPI.

### TCAPI V3 send path
* Adds tcapi.SendArgs, tcapi.RunConfig, and tcapi.SendV3Message — pair-level V3 encode/send via BuildV3ExtraArgs → BuildChainMessage → SendChainMessage instead of deprecated Chain.SendMessage.
* Introduces basic.Args and token_transfer.Args (Send, Run, plus token-only TransferAmount / DestBalanceIncrease for cross-family balance checks). Zero values preserve prior EVM→EVM defaults.
* Breaking: basic.All, all exported basic.* constructors, token_transfer.All / All17, and token_transfer.TokenTransfer now take Args (or basic.Args) instead of bare tcapi.SendArgs / no config.
* Confirm path now keys off MessageID from SendV3Message (no pre-send GetExpectedNextSequenceNumber), matching composable messaging tests.

### Chain-agnostic contract resolution
* Adds chainreg.AddressResolver (receiver, executor, committee CCV, token) registered per family in chainreg.Registry.
* Removes tcapi.GetContractAddress and tcapi/contracts.go. TCAPI hydration uses loadV3Env + per-family resolvers; cases skip when a family lacks AddressResolver.

### Token address lookup (pool → token)
* AddressResolver.GetToken(ds, chainSelector, poolRef) resolves the token contract for balance checks, not the pool address. Callers pass the full pool datastore.AddressRef; each family maps pool type/version/qualifier → token ref.
* EVM: new evm.TokenRefForPool (build/devenv/evm/tokens.go) — generic BurnMint/LockRelease pools → BurnMintERC20WithDripToken v1.0.0; Lombard qualifier → v1.5.0. Shared with buildEVMTokenTransferConfig in evm/impl.go.

## Testing
* TestE2ESmoke_Basic exercises the full tcapi/basic and tcapi/token_transfer matrices against the in-repo devenv env.
* Token transfer cases that depend on pools not deployed in the current env (e.g. LockRelease 1.6.1 on EVM-only, [default secondary] CCV mixes) are skipped via HavePrerequisites as before.

## Checklist


tt-cll (dismissed):

huangzhen1997 (dismissed):

build/devenv/common/common.go
```diff
@@ -141,7 +141,16 @@ func (s TokenCombination) FinalityConfig() protocol.Finality {
...
+// (CCV qualifier mixes, expected receipt/verification counts, etc.).
+//
+// This is NOT the same set that devenv deploys. Deployment uses ComputeTokenCombinations,
+// which derives combos from per-chain pool capabilities and ccvQualifierPermutations
+// (e.g. [], [default], [secondary], [tertiary], [default secondary tertiary] — but not
+// arbitrary subsets like [default secondary]). AllTokenCombinations also includes pool
+// pairings that no chain in the current env supports (e.g. LockRelease 1.6.1 on EVM-only
+// devenv). Smoke/TCAPI tests that iterate this list will skip cases whose pool refs are
+// absent from the datastore; align the two sources in a follow-up.
 func AllTokenCombinations() []TokenCombination {
```
makramkd, line 154: This was causing some pretty puzzling skips in the current test suite, I'll create a ticket to revisit in a follow-up.

Farber98 (approved):

winder (approved):

build/devenv/chainreg/types.go
```diff
@@ -83,6 +85,27 @@ type ExecutorModifier = executor.ReqModifier
 // ExtraArgsSerializer serializes message extra args for a destination chain family.
 type ExtraArgsSerializer = cciptestinterfaces.ExtraArgsSerializer
 
+// AddressResolver is used by the test framework to resolve addresses of certain
+// on-chain contracts. It is expected that it is implemented per-family.
+type AddressResolver interface {
```
agusaldasoro, line 90: Do we need a new interface or maybe this can be wrapped in https://github.com/smartcontractkit/chainlink-ccip/blob/main/deployment/v2_0_0/adapters/chain_family.go#L106 ?

build/devenv/common/common.go
agusaldasoro: Also, not part of this PR, but as an specific chain implementer handling this combos is a litle bit complicated to follow. A nice to have will be for the api to handle setting up one lane per time instead of handling all combos at the same time.

makramkd: This is doable today, if you look at the Solana PR we pass explicit token combos to the test case.

build/devenv/chainreg/types.go
makramkd: Ah yes, we're aware of all the different interfaces. We could potentially move these methods that just do lookups to one place; we're trying to balance having giant interfaces vs. small self-contained ones so that people can make progress w/out having to implement a lot of methods.

PR #1126, opened 2026-05-20
CI: spin up envs using production images

makramkd:
## Description

So far we've been testing with the "dev" images which provide some hot
reload functionality which is in theory useful in dev mode. There are
downsides to this though:

1. The dev image is not the one shipped to producation. This means we're
   currently only testing dev images in CI, which are built using
   a different Dockerfile, which can diverge from the production
   version.
2. The dev image is far slower to start because it recompiles the Go
   code on each run. This is wasted cycles in CI because we spin up way
   more containers than we build images.

This commit:

1. Updates then env*.toml files to use ":latest" instead of ":dev",
   which are tags that refer to the images built using the production
   Dockerfiles.
2. Updates CI to test using the production images.

## Checklist

### DIFF
--- a/build/devenv/tests/e2e/aggregatorcli/client.go
+++ b/build/devenv/tests/e2e/aggregatorcli/client.go
@@ -18,7 +18,6 @@ import (
 
 const (
 	// DefaultBinaryPath is the in-container path of the aggregator binary.
-	// In devenv the binary is built by air into /tmp/aggregator (see aggregator/air.toml).
 	DefaultBinaryPath = "/bin/aggregator"
 )
 
--- a/build/devenv/tests/e2e/aggregatorcli/client.go
+++ b/build/devenv/tests/e2e/aggregatorcli/client.go
@@ -19,7 +19,7 @@ import (
 const (
 	// DefaultBinaryPath is the in-container path of the aggregator binary.
 	// In devenv the binary is built by air into /tmp/aggregator (see aggregator/air.toml).
-	DefaultBinaryPath = "/tmp/aggregator"
+	DefaultBinaryPath = "/bin/aggregator"
 )
 
 // Client talks to a single aggregator container. It is cheap to construct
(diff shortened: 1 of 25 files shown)


build/devenv/tests/e2e/chaos_test.go
```diff
@@ -110,7 +110,7 @@ func TestChaos_VerifierFaultToleranceThresholdViolated(t *testing.T) {
 	containerRe2 := fmt.Sprintf("(%s)", strings.Join(func() []string {
 		names := make([]string, 0, len(toStop))
 		for _, verifier := range toStop {
-			names = append(names, verifier.Out.ContainerName)
+			names = append(names, fmt.Sprintf("^%s$", verifier.Out.ContainerName))
```
makramkd, line 113: This was causing flakes because it would stop the DB and the app container, causing a race on startup. The apps should probably retry on startup to reach the DB rather than fail completely after one try. Will create a ticket for this.

build/devenv/tests/e2e/chaos_test.go
```diff
@@ -164,7 +164,7 @@ func TestChaos_AllExecutorsDown(t *testing.T) {
 	var defaultExecutorContainerNames []string
 	for _, executor := range setup.in.Executor {
 		if executor.ExecutorQualifier == devenvcommon.DefaultExecutorQualifier {
-			defaultExecutorContainerNames = append(defaultExecutorContainerNames, executor.Out.ContainerName)
+			defaultExecutorContainerNames = append(defaultExecutorContainerNames, fmt.Sprintf("^%s$", executor.Out.ContainerName))
```
makramkd, line 167: Ditto here.

build/devenv/tests/e2e/aggregatorcli/client.go
```diff
@@ -18,8 +18,7 @@ import (
 
 const (
 	// DefaultBinaryPath is the in-container path of the aggregator binary.
-	// In devenv the binary is built by air into /tmp/aggregator (see aggregator/air.toml).
-	DefaultBinaryPath = "/tmp/aggregator"
+	DefaultBinaryPath = "/bin/aggregator"
```
winder, line 21: Ideally this sort of thing is a property of the docker image and the code doesn't know or care about it. [Ticket created.](https://smartcontract-it.atlassian.net/browse/CCIP-11512)

winder (approved):

build/devenv/tests/e2e/chaos_test.go
makramkd: [Ticket](https://smartcontract-it.atlassian.net/browse/CCIP-11514)

jadepark-dev (approved):

PR #1132, opened 2026-05-21
Zero address is allowed for receiver

huangzhen1997:
Loosen the check for receiver for the case of Solana token transfer https://docs.chain.link/ccip/tutorials/svm/destination/token-transfers#understanding-token-transfers-to-svm

### DIFF
--- a/verifier/pkg/commit/validation.go
+++ b/verifier/pkg/commit/validation.go
@@ -22,6 +22,10 @@ func ValidateVerificationTask(verificationTask *verifier.VerificationTask) error
 		return fmt.Errorf("sender cannot be empty or zero")
 	}
 
+	if len(message.Receiver) == 0 {
+		return fmt.Errorf("receiver cannot be empty")
+	}
+
 	if len(verificationTask.ReceiptBlobs) == 0 {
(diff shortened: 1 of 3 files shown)


verifier/pkg/commit/validation.go
```diff
@@ -22,8 +22,8 @@ func ValidateVerificationTask(verificationTask *verifier.VerificationTask) error
 		return fmt.Errorf("sender cannot be empty or zero")
 	}
 
-	if message.Receiver.IsZeroOrEmpty() {
-		return fmt.Errorf("receiver cannot be empty or zero")
+	if len(message.Receiver) == 0 {
+		return fmt.Errorf("receiver cannot be empty")
```
jadepark-dev: tiny nit: can be a follow-up, I would suggest to implement the same pattern. Also would be great if we could add one liner comment on zero receiver use case
```
func (a UnknownAddress) IsEmpty() bool {
```

jadepark-dev (dismissed):

RensR (approved):

KodeyThomas (approved):

PR #1134, opened 2026-05-22
extract LoadGun interface and AssertMessagesAsync into load

Farber98:
## Description

Introduces a `LoadGun` interface and moves `AssertMessagesAsync` into `tests/e2e/load/` so non-EVM chains can reuse the shared verification and metrics pipeline without importing go-ethereum.

This PR breaks that coupling by:
- Defining a `LoadGun` interface in `load/` that any chain-specific gun can implement
- Moving the verification pipeline (`AssertMessagesAsync`) into the same package
- Introducing `VerificationContext` to avoid a circular import between `load/` and `e2e/`
- Replacing `common.BytesToHash` with `encoding/hex` to remove the go-ethereum dep

After this change, `load/` depends only on stdlib, wasp, cciptestinterfaces, and metrics. No EVM deps. Solana (and future chains) implement `LoadGun`, construct a `VerificationContext`, and call `load.AssertMessagesAsync` directly.

No behavioral changes to existing EVM load tests.

### DIFF
--- a/build/devenv/tests/e2e/load/assert_messages.go
+++ b/build/devenv/tests/e2e/load/assert_messages.go
@@ -39,15 +39,20 @@ func AssertMessagesAsync(vc VerificationContext, gun LoadGun, overallTimeout tim
 	go func() {
 		defer cancelVerify()
 
+		logTimeout := sync.OnceFunc(func() {
+			vc.T.Logf("Overall verification timeout reached, stopping new verifications")
+		})
+
 		for sentMsg := range gun.SentMessages() {
 			select {
 			case <-verifyCtx.Done():
-				vc.T.Logf("Overall verification timeout reached, stopping new verifications")
-				return
+				// Log once and keep draining to avoid blocking the gun's producer goroutine.
+				logTimeout()
+				continue
 			default:
 			}
 
-			msgIDHex := hex.EncodeToString(sentMsg.MessageID[:])
+			msgIDHex := "0x" + hex.EncodeToString(sentMsg.MessageID[:])
 			totalSent.Add(1)
 			sentMessages.Store(sentMsg.SeqNo, msgIDHex)
 
@@ -55,7 +60,7 @@ func AssertMessagesAsync(vc VerificationContext, gun LoadGun, overallTimeout tim
 			go func(msg SentMessage) {
 				defer wg.Done()
 
-				msgIDHex := hex.EncodeToString(msg.MessageID[:])
+				msgIDHex := "0x" + hex.EncodeToString(msg.MessageID[:])
 
 				if _, ok := vc.Impl[msg.ChainPair.Dest]; !ok {
 					vc.T.Logf("No implementation available to verify message %d", msg.SeqNo)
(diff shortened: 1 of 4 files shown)


build/devenv/tests/e2e/load/assert_messages.go
```diff
@@ -0,0 +1,156 @@
...
+		logTimeout := sync.OnceFunc(func() {
+			vc.T.Logf("Overall verification timeout reached, stopping new verifications")
+		})
+
+		for sentMsg := range gun.SentMessages() {
+			// Always record sent messages for accurate totals, even after timeout.
+			msgIDHex := "0x" + hex.EncodeToString(sentMsg.MessageID[:])
+			totalSent.Add(1)
+			sentMessages.Store(sentMsg.SeqNo, msgIDHex)
+
```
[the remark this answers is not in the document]
Farber98: guns are single lane for now

build/devenv/tests/e2e/load/assert_messages.go
```diff
@@ -0,0 +1,156 @@
...
+	Ctx  context.Context
+	T    testing.TB
+	Impl map[uint64]cciptestinterfaces.CCIP17
+}
+
+// AssertMessagesAsync starts a background verification pipeline that reads
+// sent messages from the gun's channel and confirms execution on dest chains.
+// Returns a closure that blocks until verification completes and returns metrics.
+// The caller must call gun.CloseSentChannel() after the load run completes to unblock the pipeline.
+func AssertMessagesAsync(vc VerificationContext, gun LoadGun, overallTimeout time.Duration) func() ([]metrics.MessageMetrics, metrics.MessageTotals) {
```
jadepark-dev: Here the parameter `gun` is only used for getting sent messages. If you change the pipeline function to accept a plain Go channel of events instead of a `LoadGun`, it becomes a pure cross-chain pipeline verifier, completely freed from the concept of load testing or generators

jadepark-dev: What do you think about moving the verification pipeline directly into the `cciptestinterfaces` package instead of adding a separate abstraction layer inside load?

Right now, `AssertMessagesAsync` is tightly coupled to the LoadGun interface just to read from a channel. Since its main job is checking if a message is verified, it feels like it might fit better as a core protocol testing tool. If we separate load generation from lifecycle verification, the pipeline could become a pure, chain-agnostic event consumer that only depends on `ConfirmExecOnDest`.

Farber98: Good catch! I'll add the first one 😃 

About the second one, let me think more about it and I'll come back

huangzhen1997: From my opinion, it makes more sense to stay here to reduce duplicate code

Farber98: Changed `AssertMessagesAsync` to accept `<-chan SentMessage` instead of `LoadGun` and it's now a pure channel consumer, freed from the load gun concept as Jade suggested 🙌

Farber98: On moving to a separate package, I explored a few alternatives:

- **Directly in `cciptestinterfaces`**: The package is currently pure types/interfaces. Adding implementation logic would change its contract.
- **Sub-package `cciptestinterfaces/verification/`**: Technically feasible, but the parent package name is "interfaces" and adding implementation children contradicts that as mentioned above. We could rename `cciptestinterfaces` → `cciptest` to make it work, but would like more opinions since the package is used everywhere and it may be changing its contract from "pure types" to "SDK"
- **Place it in `tests/e2e/verification/`**: Follows the existing pattern of sibling packages under `tests/e2e/` (like `logasserter/`, `metrics/`). 

For me, most viable path forward if we want to move would be option 3. Also fine as keeping as it is. 

Would love to hear your thoughts on this one guys!

winder: I'm fine with the current implementation or option 3. Keeping these interfaces separate will help hopefully encourage isolation, which is important for the TDD / Test Suite approach to integrations that we've been discussing.

jadepark-dev (approved):

build/devenv/tests/e2e/load/assert_messages.go
Farber98: Nice! Let's keep current impl and we can revisit the package org later if needed. Solana impl may give us another perspective

winder (approved):

PR #1139, opened 2026-05-29
fix: Increase fee limit to compensate for EIP-150 gas attenuation

nvsriram:
## Description

The `executionGasLimit` is currently used as the destination transaction fee limit. However, it does not account for EIP-150 forwarding at most ~63/64 gas with each external call. 

This PR compensates for this attenuation by adding a buffer ($`\left\lceil{\frac{64}{63} \left\lceil{\frac{64}{63}L}\right\rceil}\right\rceil - L`$, where $`L`$ is the `ccipReceiveGasLimit`) to account for the two external calls made by the OffRamp (Router -> Receiver).

## Testing

- Updated contract transmitter tests to verify fee limit includes this new buffer
- Added coverage where `executionGasLimit` and `ccipReceiveGasLimit` differ

### DIFF
--- a/integration/pkg/contracttransmitter/txm_evm_contract_transmitter.go
+++ b/integration/pkg/contracttransmitter/txm_evm_contract_transmitter.go
@@ -20,10 +20,10 @@ import (
 )
 
 const (
-	// Numerator for EIP-150 compensation across two CALL boundaries: 64^2.
-	eip150ForwardingNumerator   = 64 * 64
-	// Denominator for EIP-150 compensation across two CALL boundaries: 63^2.
-	eip150ForwardingDenominator = 63 * 63
+	// Numerator for EIP-150 gas buffer across two CALL boundaries: 64^2 - 63^2.
+	eip150ForwardingBufferNumerator = 64*64 - 63*63
+	// Denominator for EIP-150 gas buffer across two CALL boundaries: 63^2.
+	eip150ForwardingBufferDenominator = 63 * 63
 )
 
 var (
@@ -77,7 +77,8 @@ func (ct *TXMEVMContractTransmitter) ConvertAndWriteMessageToChain(ctx context.C
 		return fmt.Errorf("skipping transmit, error getting round-robin from address: %w", err)
 	}
 	messageID, _ := report.Message.MessageID()
-	feeLimit := eip150ForwardingFeeLimit(report.Message.ExecutionGasLimit)
+	feeLimit := uint64(report.Message.ExecutionGasLimit) +
+		eip150ForwardingGasBuffer(report.Message.CcipReceiveGasLimit)
 
 	// we don't want to use an idempotency key based on messageid in case the CCV Data changes in between resubmissions
 	tx, err := ct.TxmClient.CreateTransaction(ctx, txmgr.TxRequest{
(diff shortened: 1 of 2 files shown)


integration/pkg/contracttransmitter/txm_evm_contract_transmitter.go
```diff
@@ -87,3 +96,11 @@ func (ct *TXMEVMContractTransmitter) ConvertAndWriteMessageToChain(ctx context.C
 	ct.lggr.Infow("submitted tx to txm", "messageID", messageID, "txm key", tx.IdempotencyKey)
 	return nil
 }
+
+// eip150ForwardingGasBuffer returns the extra gas needed to compensate for EIP-150
+// gas attenuation across the OffRamp -> Router -> Receiver call path.
+// It computes ceil(((64^2 - 63^2) / 63^2) * L), where L is the ccipReceive gas limit.
+func eip150ForwardingGasBuffer(ccipReceiveGasLimit uint32) uint64 {
```
RensR, line 102: Could we add a simple unit tests that asserts something like this? 
```
ccipReceiveGasLimit := 200_000
withBuffer := eip150ForwardingGasBuffer(ccipReceiveGasLimit)

assert(floor(withBuffer / 64 * 63 / 64 * 63),  ccipReceiveGasLimit)
```

All of the current tests don't test this code, just that the function is being called.

nvsriram: Done!

integration/pkg/contracttransmitter/txm_evm_contract_transmitter.go
```diff
@@ -87,3 +95,18 @@ func (ct *TXMEVMContractTransmitter) ConvertAndWriteMessageToChain(ctx context.C
 	ct.lggr.Infow("submitted tx to txm", "messageID", messageID, "txm key", tx.IdempotencyKey)
 	return nil
 }
+
+// eip150ForwardingGasBuffer returns the extra gas needed to compensate for EIP-150
+// gas attenuation across the OffRamp -> Router -> Receiver call path.
+// It computes ceil(64/63 * ceil(64/63 * L)) - L, where L is the ccipReceive gas limit.
```
crispymangoes, line 101: The math is mathing! The only thing this is missing is technically when the router makes the receiver call it enforces that we have an extra 5k gas(configured in `OffRamp::i_gasForCallExactCheck`), so the most accurate number would be taking what you have now and adding 5k to it. 

Though if txs are not going through because we are missing this 5k, then I think the real problem is that we are exact changing the gas used vs what is allotted, which means we should update our gas values on chain. @RensR

nvsriram: I see, that makes sense. The only catch with adding the 5k would be that the offRamp config isn't currently stored in the transmitter. So that limits options to:
1. Read/store the config from on-chain to access `i_gasForCallExactCheck`
2. Store 5k as a constant and remember to update it if `i_gasForCallExactCheck` changes

RensR: This value should already be coverer by the destGasOverhead. Imo we can ignore it here. The 1/64 of that is basically nothing, even doing it twice. It's also constant so we can still keep it in destGasOverhead.

RensR (approved):

KodeyThomas (approved):

PR #1138, opened 2026-05-29
build/devenv: toml config updates

winder:
# PR Chain
There are multiple open pull requests that feed into each other:

- https://github.com/smartcontractkit/chainlink-ccv/pull/1148
- https://github.com/smartcontractkit/chainlink-ccv/pull/1145
- https://github.com/smartcontractkit/chainlink-ccv/pull/1143
- https://github.com/smartcontractkit/chainlink-ccv/pull/1141
- https://github.com/smartcontractkit/chainlink-ccv/pull/1140
- **this PR** https://github.com/smartcontractkit/chainlink-ccv/pull/1138
- https://github.com/smartcontractkit/chainlink-ccv/pull/1137
- https://github.com/smartcontractkit/chainlink-ccv/pull/1133
- main

## Description
- Added version field to phased runtime config — top-level and per-component; validated on startup, missing or mismatched versions error
- Split env.toml into env.toml (monolith) and env-phased.toml (phased) — the two modes have incompatible config layouts
- Unified committeeccv config under [committeeccv] and topology under [protocol_contracts.environment_topology]
- Phased CI smoke test now boots from env-phased.toml

## Checklist

- [ ] Breaking changes documented in changelog (see `changelog` directory)
- [ ] Cross link related PRs (in this or other repositories)
- [ ] `just lint fix` - no new lint errors
- [ ] `just generate` - mocks and protobufs are up to date

PR #1137, opened 2026-05-29
build/devenv: remove fallback, scope lane flag, merge executor phases.

winder:
# PR Chain
There are multiple open pull requests that feed into each other:

- https://github.com/smartcontractkit/chainlink-ccv/pull/1148
- https://github.com/smartcontractkit/chainlink-ccv/pull/1145
- https://github.com/smartcontractkit/chainlink-ccv/pull/1143
- https://github.com/smartcontractkit/chainlink-ccv/pull/1141
- https://github.com/smartcontractkit/chainlink-ccv/pull/1140
- https://github.com/smartcontractkit/chainlink-ccv/pull/1138
- **this PR** https://github.com/smartcontractkit/chainlink-ccv/pull/1137
- https://github.com/smartcontractkit/chainlink-ccv/pull/1133
- main

## Description
Phased devenv cleanup: Four independent follow-up cleanups now that the legacy fallback component has been fully eliminated from the phased devenv runtime.

### Remove `SetFallback` from the runtime registry

  `SetFallback` / `r.fallback` is dead code — nothing registers a
  fallback component any more. This removes the field from `Registry`,
  both `SetFallback` functions (instance + package-level), the
  `fallbackOwner` constant, and all four per-phase fallback execution
  branches in `environment.go`. The two fallback-specific unit tests are
  also deleted.

### Stop routing `use_legacy_configure_lane` through an inter-phase key

  `protocol_contracts` previously read the flag from its component config
  and re-exported it as `"_use_legacy_configure_lane"` so that
  `committeeccv` could consume it one phase later. Now that both
  components share `globalConfig`, `committeeccv` reads the value
  directly from `globalConfig["protocol_contracts"]`. The `protocol_contracts`
  component no longer touches the flag at all. No TOML changes — the
  field stays at `[protocol_contracts].use_legacy_configure_lane`.

### Relax `NodeSets` validation tag

  `Cfg.NodeSets` had `validate:"required"`, which fails for phased
  configs that have no `[[nodeset]]` sections. Changed to
  `validate:"omitempty"`. Monolith configs always pass `env-cl.toml`
  which supplies `NodeSets`, so the monolith path is unaffected.

### Merge executor into a single Phase 3 component

  Executor was split across Phase 3 (launch + JD registration) and Phase
  4 (job spec generation). The split was necessary when
  `protocol_contracts` ran in Phase 3 — job specs needed `_env`,
  `_topology`, and `_ds` which weren't available until Phase 3 completed.
[...]

PR #1133, opened 2026-05-21
phased devenv: finish components and remove fallback

winder:
# PR Chain
There are multiple open pull requests that feed into each other:

- https://github.com/smartcontractkit/chainlink-ccv/pull/1148
- https://github.com/smartcontractkit/chainlink-ccv/pull/1145
- https://github.com/smartcontractkit/chainlink-ccv/pull/1143
- https://github.com/smartcontractkit/chainlink-ccv/pull/1141
- https://github.com/smartcontractkit/chainlink-ccv/pull/1140
- https://github.com/smartcontractkit/chainlink-ccv/pull/1138
- https://github.com/smartcontractkit/chainlink-ccv/pull/1137
- **this PR** https://github.com/smartcontractkit/chainlink-ccv/pull/1133
- main

## Description
This PR continues work on the phased devenv to move all logic into components. The legacy fallback component has been removed.

* Token Verifier Component: added the final new component.
* Drop CL node support: this had odd dependencies with the committeeCCV and is not part of the long term vision. If needed, it can be added back in with a future PR.
* CommitteeCCV Component: absorbed missing business logic (HMAC credential generation, verifier service management, TLS certificate logic, aggregator service management).
* Executor Component: absorbed launch logic, it now has logic in multiple phases (to be addressed in future PR)

## Testing
No new functionality, existing regression tests pass.

CI was temporarily updated to run all smoke tests with the phased environment to ensure more complete regression test coverage.

## Checklist

- [ ] Breaking changes documented in changelog (see `changelog` directory)
- [ ] Cross link related PRs (in this or other repositories)
- [ ] `just lint fix` - no new lint errors
- [ ] `just generate` - mocks and protobufs are up to date

PR #1140, opened 2026-05-29
build/devenv: serialize raw phased output map, version-aware load

winder:
# PR Chain
There are multiple open pull requests that feed into each other:

- https://github.com/smartcontractkit/chainlink-ccv/pull/1148
- https://github.com/smartcontractkit/chainlink-ccv/pull/1145
- https://github.com/smartcontractkit/chainlink-ccv/pull/1143
- https://github.com/smartcontractkit/chainlink-ccv/pull/1141
- **this PR** https://github.com/smartcontractkit/chainlink-ccv/pull/1140
- https://github.com/smartcontractkit/chainlink-ccv/pull/1138
- https://github.com/smartcontractkit/chainlink-ccv/pull/1137
- https://github.com/smartcontractkit/chainlink-ccv/pull/1133
- main

## Description
The phased devenv (--env-mode phased) previously reconstructed a transitional *Cfg from the component runtime's output map via hand-written sync blocks. This PR removes that bridge: NewPhasedEnvironment now serializes the raw accumulated output map directly, and LoadOutput[Cfg] switches on a version marker to decode either format. Adding a new serialized output no longer requires touching a central mapping — a component just publishes a public key.

### What changed

- NewPhasedEnvironment returns the raw accumulated map and writes the output file via storePhasedOutput, which strips runtime-only keys (any key prefixed with _) and serializes the rest. The schema version is re-published as a public key so the file starts with version = N.
- Component output keys reclassified into public (serialized) vs. _-private (runtime-only, stripped):
    - protocol_contracts: _cldf → cldf, _topology → environment_topology (consumers in executor/committeeccv updated).
    - committeeccv: shared_tls_certs → _shared_tls_certs (consumer in indexer updated).
    - jd stays public (needed for future job proposals); JDInfrastructure.OffchainClient (a live gRPC client) is tagged toml:"-" so the section serializes cleanly while keeping jd_output + node_id_map.
- LoadOutput[Cfg] is now version-aware — a single dispatch point that all readers (NewLibFromCCVEnv and the ~18 direct LoadOutput[ccv.Cfg] test call sites) funnel through:
    - version absent/0 → existing strict Cfg decode (legacy/monolith).
    - version == 1 → lenient decode of the raw map + derive aggregator/indexer endpoint maps from each launched service's Out.
    - any other version → error.
- env.toml drops its version key so monolith output stays at 0 and routes to the legacy decoder.

### Why

  - Removes the fragile, hand-maintained *Cfg reconstruction in the phased path.
  - Establishes a clear "public key vs. _-private runtime key" convention owned by each component.
  - Keeps the output-file contract backward compatible: no reader signatures or call sites change, because the version dispatch lives inside LoadOutput.

### Compatibility / risk

  - Legacy (monolith) output and all its readers are unchanged.
  - Phased output keys are a raw component dump; the phased decode is intentionally lenient and ignores keys Cfg doesn't model (e.g. the aggregators/verifiers plurals).

## Checklist

### DIFF
--- a/build/devenv/config.go
+++ b/build/devenv/config.go
@@ -26,6 +26,8 @@ import (
 	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
 
 	devenvcommon "github.com/smartcontractkit/chainlink-ccv/build/devenv/common"
+	"github.com/smartcontractkit/chainlink-ccv/build/devenv/services"
+	"github.com/smartcontractkit/chainlink-ccv/build/devenv/services/committeeverifier"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


build/devenv/config.go
```diff
@@ -93,23 +95,47 @@ func Store[T any](cfg *T) error {
...
 }
 
-// LoadOutput loads config output file from path.
-// TODO: why is this generic?
+// LoadOutput loads a devenv output file and returns a populated *Cfg. It
+// supports two output formats, distinguished by the top-level "version" key:
+//
+//   - version 0 / absent — the legacy monolith output, a strict Cfg dump.
+//   - version 1 — the phased output, a raw component-output map; the required
+//     components are decoded out of it and the endpoint maps are derived.
```
makramkd, line 103: **question**: can these be made constants defined somewhere in the `devenv` package?

makramkd (dismissed):

makramkd (approved):

build/devenv/config.go
winder: I don't think it's useful to treat a version number like a feature flag. This is more of a backwards compatibility gate. If we end up needing a constant at some point we can add it later.

build/devenv/config.go
```diff
@@ -93,23 +95,47 @@ func Store[T any](cfg *T) error {
...
+		return nil, fmt.Errorf("failed to read config version from %s: %w", outputPath, err)
 	}
+
+	var c *Cfg
+	switch probe.Version {
+	case 0:
+		c, err = loadLegacyCfg(data)
+	case 1:
+		c, err = loadPhasedCfg(data)
+	default:
```
winder, line 127: I see why you might want a constant, but this (hopefully) goes away eventually. If we keep legacy around for an extended duration this probably changes to "version == 0 { legacy } else { phased }"

huangzhen1997 (approved):

PR #1155, opened 2026-06-05
Resolve transmitter keys via chainreg instead of per-family keys

ogtownsend:
## Description
Adds Solana executor signer support to devenv and removes the per chain family transmitter wiring 

This PR moves bootstrap keys into the `chainreg` registry:
- `BootstrapKeys` now exposes a generic `PublicKeys map[string]string` (keyed by full keystore key name) plus a `PublicKeyHex(keyName)` accessor, instead of one named field per family. `FetchBootstrapKeys` populates it generically and no longer derives a family-specific address itself.
- New `chainreg.ExecutorInfo` interface (`ExecutorTransmitterKeyName()`, `ExecutorTransmitterAddress(keys)`) registered via `Registration.ExecutorInfo`, with `Registry.GetExecutorTransmitterKeyName` / `GetExecutorTransmitterAddress` lookups. EVM registers its own `ExecutorInfo`. 
- New families (Solana, stellar, etc.) register theirs from their own repo so there shouldn't be any need for shared devenv edits

## Testing
- `go build ./...` and `cd build/devenv && go build ./...` both pass.
- `go test ./chainreg/...` passes, including new `TestRegistryGetExecutorTransmitterInfo` and `TestRegistryAddMergesExecutorInfo` covering the registry lookups and merge.
- Tested with Solana devenv and new transmitter funding logic, sending messages EVM<-->Solana: https://github.com/smartcontractkit/chainlink-ccip-solana/pull/201 

## Checklist


build/devenv/services/bootstrap.go
```diff
@@ -101,6 +101,9 @@ type BootstrapKeys struct {
 	// EVMTransmitterAddress is the Ethereum address derived from the executor's EVM transmitter key.
 	// Only populated for executor nodes.
 	EVMTransmitterAddress string `toml:"evm_transmitter_address,omitempty"`
+	// SolanaTransmitterAddress is the hex-encoded 32-byte Ed25519 public key of the executor's
+	// Solana transmitter key. Only populated for Solana-family executor nodes.
+	SolanaTransmitterAddress string `toml:"solana_transmitter_address,omitempty"`
```
makramkd: **suggestion**: I feel like this should probably be a mapping so that every new chain family doesn't have to update the devenv w/ their family's transmitter key, wdyt?

executor/const.go
```diff
@@ -10,4 +10,11 @@ const (
 	// devenv uses it to look up the funded on-chain address after the container
 	// is running.
 	DefaultEVMTransmitterKeyName = "evm/tx/executor_evm_transmitter_key"
+
+	// DefaultSolanaTransmitterKeyName is the full keystore path of the Ed25519 key used
+	// to sign and submit OffRamp transactions on Solana chains. The solana/tx/ prefix
+	// follows the solkeys convention so that TxKey can use it directly with WithNoPrefix().
+	// The bootstrap framework creates or loads it on startup; devenv uses it to look up
+	// the funded on-chain address after the container is running.
+	DefaultSolanaTransmitterKeyName = "solana/tx/executor_solana_transmitter_key"
```
makramkd: **suggestion**: I believe this is an anti-pattern, IMO DefaultEVMTransmitterKeyName should not have been declared in the top-level `executor` package but an `evm` package or similar.

build/devenv/chainreg/types.go
```diff
@@ -51,6 +51,19 @@ type ImplFactory interface {
 	// funding of executor addresses. Families that lack on-chain transfer
 	// primitives in devenv (e.g. Canton) return false.
 	SupportsFunding() bool
+
+	// ExecutorTransmitterKeyName returns the keystore key name that the executor
+	// for this chain family declares (via bootstrap.WithKey) and that devenv must
+	// fetch from the bootstrap server to learn the on-chain transmitter address.
+	// Return "" if the family has no bootstrap-managed transmitter key.
+	ExecutorTransmitterKeyName() string
```
makramkd, line 64: **suggestion**: I wonder if this is better directly registered into the chainreg.Registry as its own thing, e.g. update `Registration` to add a new `ExecutorTransmitterKeyName` field. Then chain families can inject the constant when calling the top-level `Register` call.

makramkd: **aside**: I'm not sure of the future of `ImplFactory`; I feel like having some modularity here would be nice though.

build/devenv/chainreg/types.go
```diff
@@ -51,6 +51,19 @@ type ImplFactory interface {
...
+
+	// ExecutorTransmitterAddress returns the executor's on-chain transmitter
+	// address (hex-encoded) for this chain family, extracted from the bootstrap
+	// keys. Each family selects the appropriate field (e.g. EVM uses
+	// EVMTransmitterAddress, Solana uses SolanaTransmitterAddress).
+	// Return "" if no transmitter address is available.
+	ExecutorTransmitterAddress(keys services.BootstrapKeys) string
```
makramkd, line 70: **suggestion**: maybe this and the other method should be its own interface, `ExecutorInfo` or something:

```golang
type ExecutorInfo interface {
  // add docstring from above
  ExecutorTransmitterKeyName() string
  // add docstring from above
  ExecutorTransmitterAddress(keys services.BootstrapKeys) string
}
```

The win here is that its a single-responsibility/purpose interface that can be optionally implemented by families as they go through the integration.

build/devenv/chainreg/types.go
makramkd: re: initial suggestion above, see https://github.com/smartcontractkit/chainlink-ccv/pull/1155#discussion_r3389881896 as an updated suggestion.

build/devenv/chainreg/types.go
ogtownsend: mmm, yes I like this

build/devenv/chainreg/registry.go
```diff
@@ -167,6 +169,49 @@ func (r *Registry) GetExecutorModifiers() map[string]ExecutorModifier {
...
+		family = chainsel.FamilyEVM
+	}
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	reg, ok := r.registrations[family]
+	if !ok || reg.ImplFactory == nil {
+		return ""
+	}
+	ei, ok := reg.ImplFactory.(ExecutorInfo)
```
makramkd: **suggestion**: I was thinking you could just add a new field to `Registration` which is `ExecutorInfo`, which can get registered as its own thing, no cast needed.

executor/const.go
ogtownsend: yeah agreed, I'll find a better place for these

build/devenv/chainreg/registry.go
ogtownsend: oh woops gotcha, misunderstood your other comment. Yeah this is a bit cleaner

build/devenv/chainreg/registry.go
```diff
@@ -167,6 +172,41 @@ func (r *Registry) GetExecutorModifiers() map[string]ExecutorModifier {
...
+	if !ok || reg.ExecutorInfo == nil {
+		return ""
+	}
+	return reg.ExecutorInfo.ExecutorTransmitterKeyName()
+}
+
+// GetExecutorTransmitterAddress returns the executor's on-chain transmitter
+// address for the given chain family, or "" if the family is unregistered or has
+// no ExecutorInfo.
+func (r *Registry) GetExecutorTransmitterAddress(family string, keys services.BootstrapKeys) string {
```
makramkd: **suggestion**: in an effort to make the import tree less wonky and to keep `Registry` as just a general store for the `Registration` object, can we do this instead:

```golang
registration, err := chainreg.GetRegistry().Get(family)
if err != nil { ... }
if registration.ExecutorInfo == nil {
  ... // use default or error, depending on the context
}
// use registration.ExecutorInfo
```

we can then get rid of `GetExecutorTransmitterAddress` and `GetExecutorTransmitterKeyName`. This way handling the presence / lack thereof of `ExecutorInfo` is up to the caller, i.e. whether to error and return, continue, etc.

build/devenv/services/bootstrap.go
```diff
@@ -98,9 +97,17 @@ type BootstrapKeys struct {
 	ECDSAPublicKey string `toml:"ecdsa_public_key,omitempty"`
 	// ECDSAAddress is the Ethereum address derived from the ECDSA public key (verifier only).
 	ECDSAAddress string `toml:"ecdsa_address,omitempty"`
-	// EVMTransmitterAddress is the Ethereum address derived from the executor's EVM transmitter key.
-	// Only populated for executor nodes.
-	EVMTransmitterAddress string `toml:"evm_transmitter_address,omitempty"`
+	// PublicKeys is a map of full keystore key names (like evm/tx/executor_evm_transmitter_key).
+	// Values are the corresponding public keys as lowercase hex-encoded raw bytes
+	PublicKeys map[string]string `toml:"public_keys,omitempty"`
```
winder, line 102: nice 👍 

cc @huangzhen1997 @faisal-chainlink

winder: *question*: Did you consider making this a `map[family]map[keyname]key`? I understand that the keystore way is to encode metadata into the key name, is that how you decided on this approach?

winder (dismissed):

build/devenv/services/bootstrap.go
faisal-chainlink: I'm curious about the same here, does this mean that an executor service must only ever specify keys that it can actually decode (i.e. matches the format that specific chain). 

In a scenario where we must specify both EVM and altVM keys w/ different key formats, this would possibly break unless explicitly handled by the `ExecutorTransmitterAddress` implementation. Handling this case was only necessary because specifying the `DefaultEVMTransmitterKeyName` key is required for a lookup that the env bootstrapping flow does a keystore lookup with it; this may not be the case anymore.

Here's an example: https://github.com/smartcontractkit/chainlink-stellar/blob/main/cmd/executor/main.go#L32-L36

ogtownsend: @winder - right, keystore already bakes it in, and because (to me) a map of maps feels like it could get unwieldy, but if there's a strong use case or benefit to quickly grabbing all keys for a chain family I'm open to it

ogtownsend: @faisal-chainlink right, after the CCV bump this workaround goes away, but you'll still need the new `ExecutorInfo` hook when you bump chainlink-ccv next in cl-stellar

build/devenv/components/executor/component.go
```diff
@@ -106,27 +111,27 @@ func (c *component) RunPhase3(
...
 			continue
 		}
 		addr, addrErr := protocol.NewUnknownAddressFromHex(addrStr)
 		if addrErr != nil {
 			return nil, nil, fmt.Errorf("executor %s invalid transmitter address: %w", exec.ContainerName, addrErr)
 		}
-		family := exec.ChainFamily
-		if family == "" {
-			family = chainsel.FamilyEVM
-		}
```
ogtownsend, line 120: FYI @winder , removing these where I can to do away with EVM special casing wherever possible, but please let me know if you'd prefer to keep them in

makramkd (approved):

winder (approved):

PR #1163, opened 2026-06-08
aggregator, indexer: reduce steady-state log verbosity and add quorum progress logs

winder:
## Description

Bring aggregator and indexer logging in line with operator targets — <50 logs at startup, <3 system logs/minute at steady state, a small bounded set per message — keeping only the per-message logs useful for tracing.

### Approach

- Demoted high-frequency diagnostic logs (per-RPC, per-heartbeat, per-aggregation-attempt, per-background-tick, per-poll) from Info to Debug, where metrics already cover steady-state observability.
- Kept one Info log per stage of the per-message lifecycle instead of every step, and tagged each with a structured `log_type` field (`message_success` / `message_failure` / `message_staso logs can be filtered and aggregated across services without parsing message text.
- **Aggregator:** added `"Verification received"` and a timer-based `"Message pending quorum"` (emitted on ther write) for in-flight visibility without spam.
- **Indexer:** collapsed the 5–7 per-attempt `Infof` calls into one structured `"Processed verification attemptt, Debug on retries); demoted the duplicate nd unified the success and DLQ paths into a single terminal `"Message processing finished"` line keyed by `status`.

### Log volume impact

**Aggregator** — idle: ~1 Info line/minute (health summary). Per message: ~N+1 Info lines (one `"Verification receivedReport submitted successfully"`), plus `"Message pending quorum"` per orphan-scan interval only while
incomplete. Scales linearly with verification ingest.

**Indexer** — idle: ~0 Info lines only). Per message: ~2V+1 Info lines (V = verifications discovered ≈ committee size), now independent of retry depth since retry passes log at Debug. The V multiplier is structural — each verifier's contribution is its own convergence task by design — not a logging artifact.

Net: steady-state noise is effectively eliminated, and per-message logging is a small, bounded, `log_type`-keyed s

## Checklist

### DIFF
--- a/protocol/log_types.go
+++ b/protocol/log_types.go
@@ -14,6 +14,12 @@ const (
 	// There is at most one of these per message on the happy path.
 	LogTypeMessageSuccess LogType = "message_success"
 
+	// LogTypeMessageFailure marks the terminal per-message failure event: a message
+	// could not be completed (e.g. it timed out and entered the dead-letter queue).
+	// Like LogTypeMessageSuccess, it is a terminal outcome — a message ends as one or
+	// the other.
+	LogTypeMessageFailure LogType = "message_failure"
+
 	// LogTypeMessageStatus marks a non-terminal per-message progress event, such as a
 	// verification being received, a message pending quorum, or a processing attempt.
 	// These can fire multiple times per message as it advances.
--- a/protocol/log_types.go
+++ b/protocol/log_types.go
@@ -1,0 +1,25 @@
+package protocol
+
+// LogType classifies a structured log line so that logs can be filtered and
+// aggregated consistently across services. Emit it under the LogTypeKey field,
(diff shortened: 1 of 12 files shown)


protocol/log_types.go
```diff
@@ -0,0 +1,31 @@
+package protocol
+
+// LogType classifies a structured log line so that logs can be filtered and
+// aggregated consistently across services. Emit it under the LogTypeKey field,
+// e.g. logger.Infow("Report submitted successfully", protocol.LogTypeKey, protocol.LogTypeMessageSuccess).
+type LogType string
+
+// LogTypeKey is the structured-log field key under which a LogType value is emitted.
+const LogTypeKey = "log_type"
```
winder, line 9: I cherry picked the most important logs to leave as Info and added a type to assist with exploring the data.

The same should be done with verifier/executor, but we probably wont have those available for launch.

anedos-chainlink (approved):

makramkd (approved):

PR #1169, opened 2026-06-10
logging, services: add WithService helper and tag root loggers

winder:
## Description

Add `common.WithService(lggr, name)` in the root `common` package and call it at each service entry point (bootstrap Run, committee verifier and executor constructors, indexer and aggregator main). Every log message emitted within a coordinator now carries a structured `"service"` field.

`WithService` lives in the top-level `common` package rather than `protocol/common/logging` because the `protocol` layer enforces a depguard rule that denies all `chainlink-common` imports; `common` is the appropriate home for shared helpers that depend on `chainlink-common/pkg/logger`.

## Checklist

### DIFF
--- a/common/logging.go
+++ b/common/logging.go
@@ -4,5 +4,6 @@ import "github.com/smartcontractkit/chainlink-common/pkg/logger"
 
 // WithService returns lggr with the "service" field set to name.
 func WithService(lggr logger.Logger, name string) logger.Logger {
-	return logger.With(lggr, "service", name)
+	// do not use the reserved "service" key.
+	return logger.With(lggr, "ccip_service", name)
 }
--- a/common/logging.go
+++ b/common/logging.go
@@ -1,0 +1,8 @@
+package common
+
+import "github.com/smartcontractkit/chainlink-common/pkg/logger"
+
+// WithService returns lggr with the "service" field set to name.
+func WithService(lggr logger.Logger, name string) logger.Logger {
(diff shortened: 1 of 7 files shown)


common/logging.go
```diff
@@ -0,0 +1,8 @@
+package common
+
+import "github.com/smartcontractkit/chainlink-common/pkg/logger"
+
+// WithService returns lggr with the "service" field set to name.
+func WithService(lggr logger.Logger, name string) logger.Logger {
+	return logger.With(lggr, "service", name)
```
makramkd: **nit(nonblocking)**: make a new const key `LogKeyService` for the `"service"` literal.

makramkd (dismissed):

common/logging.go
winder: Fair enough, but I made a function instead so that it's only used in 1 place :)

jadepark-dev (dismissed):

bukata-sa (approved):

common/logging.go
```diff
@@ -0,0 +1,9 @@
+package common
+
+import "github.com/smartcontractkit/chainlink-common/pkg/logger"
+
+// WithService returns lggr with the "service" field set to name.
```
makramkd, line 5: **nit(nonblocking)**: 
```suggestion
// WithService returns lggr with the "ccip_service" field set to name.
```

makramkd (approved):

PR #1171, opened 2026-06-12
aggregator: add strict validate-config command

makramkd:
## Description

A deploy of the aggregator recently crashed at startup because the checked-in deploy config had drifted from this repo's config struct: a `common.Duration` field (`storage.connMaxLifetime`) was still set to a bare `0`, which the duration unmarshaler rejects. Two classes of drift caused it, neither caught before deploy:

1. **Type mismatch (hard fail):** quoted-duration fields still holding bare ints, only surfaces as a runtime crash.
2. **Silently-ignored keys (soft fail):** the loaders use non-strict `toml.Decode`/`Unmarshal`, so renamed/removed keys are dropped and the app falls back to code defaults with no error (e.g. an orphan-recovery key drift meant the intended value was silently ignored).
3. **Semantic** — bad thresholds, invalid addresses, missing committee, out-of-range values → validation error

This PR adds the first piece of a pre-deploy CI check that catches both classes before they ship:

- A new `validate-config` subcommand on the aggregator binary that strictly decodes the config the same way the service loads it at startup, including the generated-config merge resolved relative to the config file , but **fails** on a decode error (type mismatch) or on any key present in the TOML but absent from the struct (`MetaData.Undecoded()`, i.e. drift).
- A shared `common/configvalidate` helper (`DecodeStrict` + `Report`) that holds the strict-decode + per-file reporting contract, so the indexer and token-verifier can reuse it as they adopt the subcommand.

**How it will be used:** the subcommand lives in the app's own runtime docker image, so it is pinned to the exact binary being deployed. A CI check in the deploy repo (`chainlink-ccv-deploy`, a separate PR) will render each env/instance's Helm config, then run `docker run <deployed-image> /bin/aggregator --config /config/aggregator.toml validate-config` against an allowlist of adopted apps, failing the PR on drift. Rollout is incremental: aggregator first (this PR), then indexer, then token-verifier in follow-ups.

## Testing

- `go test ./common/configvalidate/... ./aggregator/pkg/configuration/...`

## Checklist


carte7000 (dismissed):

common/configvalidate/configvalidate.go
makramkd, line 1: This package will be used by other services (indexer, verifier, executor) as well.

carte7000 (approved):

nvsriram (approved):

PR #1173, opened 2026-06-12
fix: update ingestion timestamp on message replay/redelivery

nvsriram:
## Description

Currently, ingestion timestamp is not updated as part of the message replay process.

This PR adds:
* Updated `gatherAllVerifications` to also upsert ingestion timestamp
* Updated `ExpirableMessageSet.dataMap` to store the latest ingestion time per messageID.

## Testing

* Added indexer replay tests to ensure upsert works as intended
* Updated executor message heap tests to validate RePush behaviour based on ingestion time
* Updated e2e smoke test to verify message ingestion timestamps have been updated as expected


executor/pkg/message_heap/message_heap.go
```diff
@@ -237,3 +248,80 @@ func (es *ExpirableMessageSet) CleanExpired(timestamp time.Time) int {
 	}
 	return expiredCount
 }
+
+// ExpirableMessageSets holds one ExpirableMessageSet per indexer source index.
+type ExpirableMessageSets struct {
```
carte7000: Can you expand on why we need to keep the message queue split by indexer? Is it because we want to avoid the different ingestion timestamp by indexer from bumping the message?

nvsriram: Yes, pretty much.

carte7000: IIRC we pick the active indexer based on its health status I am not sure how often we would end up having duplicate message coming from both indexer at once. I am wondering if we really need this additional layer in that case what do you think?

nvsriram: Yes, currently [`ReadMessages`](https://github.com/smartcontractkit/chainlink-ccv/blob/main/executor/pkg/adapter/adapter.go#L229) -> [`queryWithFailover`](https://github.com/smartcontractkit/chainlink-ccv/blob/main/executor/pkg/adapter/adapter.go#L99) returns the active indexer result on success. On failure, it queries alternate and updates the active indexer. There doesn't seem to be much use for the indexer-specific caching as of now so I'll go ahead and revert those changes.

carte7000 (approved):

KodeyThomas (approved):

makramkd (approved):

PR #1179, opened 2026-06-16
build/devenv: split protocol-contract and committee-verifier deploys

winder:
## Description

Splits the devenv's single chainlink-ccip "kitchen-sink" `DeployChainContracts`
deploy so each contract concern is deployed in the phase that owns it.

- **Phase 2 (`protocol_contracts`)** deploys only the core protocol contracts via
  the ccv `DeployProtocolContracts` changeset. Per-chain deploy tunables
  (executors, test router, EVM family extras) now come from a new
  `[protocol_contracts.deploy]` config section instead of hard-coded Go.
- **Phase 3 (`committeeccv`)** deploys committee verifiers + resolvers via the ccv
  `DeployCommitteeVerifier` changeset (before lane config), then deploys mock
  receivers via a new optional `MockReceiverDeployer` hook. Token-transfer
  configuration also moves here, since it depends on the resolver.
- The EVM `ProtocolContractsDeployAdapter` now ships in chainlink-ccip (bumped to
  `c9b0aed`) and self-registers via `init()`; the temporary in-repo wrapper is
  removed.

USDC/Lombard **pools** stay in Phase 2 — only the resolver-dependent mock receivers
move to Phase 3. `DeployerContract` (the EVM CREATE2 factory) is still resolved at
runtime via the existing `GetDeployChainContractsCfg`; no new interface method was
added.

## Testing

- `go build ./...`, `go vet`, and `just lint fix` on touched packages — clean.
- Full phased bring-up (`ccv up --env-mode phased env-phased.toml`) succeeds: Phase
  2 deploys protocol contracts only; Phase 3 deploys committee verifiers +
  resolvers + mock receivers and configures lanes.
- `TestE2ESmoke_Basic` against the phased env — all 16 subtests pass.

## Checklist

- [X] Breaking changes documented in changelog (see `changelog` directory)
- [X] Cross link related PRs (in this or other repositories)
- [X] `just lint fix` - no new lint errors
- [X] `just generate` - mocks and protobufs are up to date


build/devenv/components/protocol_contracts/component.go
```diff
@@ -156,7 +160,7 @@ func (p *component) RunPhase2(
 		}
 		chainDS := datastore.NewMemoryDataStore()
 
-		dsi, derr := ccdeploy.DeployContractsForSelector(ctx, e, impl, networkInfo.ChainSelector, topology)
+		dsi, derr := deployProtocolContractsForSelector(ctx, e, impl, networkInfo.ChainSelector, topology, cfg.Deploy)
```
winder, line 164: The main change. This local version uses a different changeset in order to skip the committee contracts.

build/devenv/evm/cctp.go
```diff
@@ -64,11 +64,9 @@ func (m *CCIP17EVMConfig) deployUSDCTokenAndPool(
...
 
-	err = m.deployCCTPMockReceivers(env, ds, selector)
-	if err != nil {
-		return fmt.Errorf("failed to deploy CCTPMockReceivers on chain %d: %w", selector, err)
-	}
-
+	// CCTP mock receivers depend on the committee-verifier resolver, which is
+	// deployed in Phase 3 (committeeccv); they are deployed there via
+	// DeployMockReceivers.
 	return nil
```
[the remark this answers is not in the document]
winder: Thanks copilot, I deleted the dead code!!!!

build/devenv/components/protocol_contracts/component.go
```diff
@@ -253,3 +274,151 @@ func decodeConfig(raw any) (config, error) {
...
+		return nil, fmt.Errorf("deployer contract not resolved for selector %d", selector)
+	}
+
+	executors, err := toCCVExecutors(deploy.Executors)
+	if err != nil {
+		return nil, fmt.Errorf("selector %d: %w", selector, err)
+	}
+
+	// 3. Deploy protocol contracts via the chain-agnostic ccv changeset.
+	out, err := ccvchangesets.DeployProtocolContracts().Apply(*env, ccvchangesets.DeployProtocolContractsInput{
```
tt-cll, line 346: this will be a new interface for altVMs and thus a breaking change, so we would want to document that in a changelog if we haven't

tt-cll: nevermind. just saw the changelog

tt-cll (approved):

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -288,6 +288,16 @@ type OnChainConfigurable interface {
...
 
+// MockReceiverDeployer is an optional interface. When implemented, devenv calls it
+// during Phase 3 (committeeccv), after committee verifiers and their resolvers have
+// been deployed, to deploy the mock receiver contracts that reference those resolvers.
+// Mock receivers can no longer be deployed in Phase 2 because they depend on the
+// committee-verifier resolver. The returned DataStore holds only the newly deployed
+// receiver addresses; families that do not implement this interface deploy no receivers.
+type MockReceiverDeployer interface {
+	DeployMockReceivers(env *deployment.Environment, selector uint64, topology *ccvdeployment.EnvironmentTopology) (datastore.DataStore, error)
+}
```
makramkd, line 299: **question**: wouldn't it be consistent to have a `ccvchangesets.DeployMockReceivers()` changeset, with a config provided by the user? That would make it config-driven from the get-go. Or is that the next step after this PR?

build/devenv/components/protocol_contracts/component.go
```diff
@@ -241,6 +241,27 @@ type config struct {
...
+	DeployTestRouter bool                      `toml:"deploy_test_router"`
+	Executors        []executorCfg             `toml:"executors"`
+	FamilyExtras     map[string]map[string]any `toml:"family_extras"`
+}
+
+// executorCfg is a single executor instance to deploy.
+type executorCfg struct {
+	Qualifier string `toml:"qualifier"`
+	Version   string `toml:"version"`
 }
```
makramkd, line 265: **question**: why don't we use `DeployProtocolContractsPerChainCfg` verbatim from `deployment/changesets`? Is there a reason for this semi-duplication?

winder: Good question, it might be a good idea except that I want to guarantee a consistent schema. Using the dependency type would risk a breaking change to the devenv config schema.

I do have some open questions and have to come back to this question in the future when I'm really scrutenizing the config schema. Specifically around how to deal with family-agnostic configuration. The upstream type is implicitly family specific (because there is a per-family implementation of `DeployProtocolContracts`). This component has an extra layer in the config for family/selector specific `FamilyExtra` overrides.

build/devenv/cciptestinterfaces/interface.go
winder: Yes. I have a followup to split this call into a separate component which will give us a place to put mock receiver configs. Whether or not it uses a separate changeset is an implementation detail.

makramkd (approved):

PR #1183, opened 2026-06-18
fancy devenv prompt

winder:
## Description
demo

## Checklist


build/devenv/reporter/bubbletea.go
makramkd, line 1: **question**: can this be used/extended for the test logs as well? For example, we currently have a consistent spam of "failed to get verifier result for messageID" until we get a success, would be nice to clean that up.

winder: No, I don't think so. Tests should use something else. The way we split test running and success checking into separate CI steps is pretty weird, I think they deserve some specific attention.

A couple options:
1. Use 'gotestsum'. It has a bunch of features for splitting up human/machine readable results. 
2. For the "failed to get verifier result for messageID" logs, we should probably improve the logging.
3. Generally, I'd like to see much more precision in the test results. The pass/fail nature of our existing assertions mean that any failure requires a full fledged investigation to figure out whats wrong.

build/devenv/cli/ccv.go
```diff
@@ -36,15 +36,20 @@ import (
...
 	LocalWASPLoadDashboard = "http://localhost:3000/d/WASPLoadTests/wasp-load-test?orgId=1&from=now-5m&to=now&refresh=5s"
 	LocalCCVDashboard      = "http://localhost:3000/d/f8a04cef-653f-46d3-86df-87c532300672/ccv-services?orgId=1&refresh=5s"
 )
 
-// newEnvFn is set by PersistentPreRunE based on the --env-mode flag.
-var newEnvFn func() error
+// newEnvFn is set by PersistentPreRunE (or applyProfile) based on the
+// --env-mode flag. The Reporter receives component-level events in phased
+// mode; legacy mode wraps the monolith with stage-level events only.
+var newEnvFn func(r devenvruntime.Reporter) error
```
winder, line 52: I wanted to ensure the TUI is fully decoupled from the devenv, thats accomplished with the Reporter interface. The environment notifies the reporter when something interesting happens. There are several reporters in this PR:
* Verbose: use the same output as today.
* Fancy: full TUI with dynamic output.
* Simple: line-based output, this is the default for non-TTY environments.

build/devenv/components/protocol_contracts/component.go
```diff
@@ -53,7 +54,23 @@ func factory(_ map[string]any) (devenvruntime.Component, error) {
...
+	p.mu.Unlock()
+}
+
+// Status implements the devenvruntime.StatusGetter optional interface so the TUI
+// reporter can poll for fine-grained progress during the long deploy loop.
+func (p *component) Status() string {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	return p.status
 }
```
winder, line 74: `Status() string` is an optional function that the runtime will call if present on the component. Many components run in a couple seconds or less, so status isn't really necessary. Others run for several minutes, this is how the runtime gets more information about what they're up to.

build/devenv/reporter/bubbletea.go
```diff
@@ -0,0 +1,381 @@
...
+	phase    int
+	name     string
+	start    time.Time
+	statuses []statusEntry
+}
+
+type tuiModel struct {
+	styles      tuiStyles
+	log         []string // completed / stage lines
+	active      map[string]*activeComp
```
winder, line 109: Currently, the runtime only support a single active component. The TUI supports having multiple.

build/devenv/cli/ccv.go
```diff
@@ -208,7 +233,11 @@ func applyEnvConfig(cmd *cobra.Command, args []string) error {
 		profilePath = positional
 	}
 	if profilePath == "" {
-		profilePath = "standard.profile"
+		if saved := getActiveConfig(); strings.HasSuffix(strings.TrimSpace(saved), ".profile") {
+			profilePath = strings.TrimSpace(saved)
+		} else {
+			profilePath = "standard.profile"
+		}
```
winder, line 240: This is a small random fix. The "active" config is stored in the home directory but we were ignoring it. 

If you run "ccv up --profile phased.profile", a subsequent call to "ccv restart" will stop the current environment and bring the phased profile up again. Without the fix it was reverting to the standard profile.

build/devenv/cli/ccv.go
```diff
@@ -706,6 +722,53 @@ func generateRunID() string {
 	return id.String()[:8]
 }
 
+// redirectToLogFile redirects os.Stdout and os.Stderr to an auto-created log
+// file in CCVConfigDir() unless verbose is true. It returns the writer that
+// the fancy reporter should render to (the saved real-terminal fd), and a
+// cleanup func that restores the original fds. In verbose mode nothing is
+// redirected and the returned writer is os.Stderr.
+func redirectToLogFile(verbose bool, prefix string) (term *os.File, cleanup func(), logPath string, err error) {
```
winder, line 730: Presumably a lot of this can be done by the logger. It would be nice to initialize the logger to a file rather than all this redirect stuff.

build/devenv/reporter/bubbletea.go
winder, line 1: You're not expected to understand this file.

I tried to hide the entirety of BubbleTEA in here. It uses "The Elm Architecture" to create an event loop for the TUI. All state changes happen through the `Update(event)` function which returns an "immutable" copy of the `tuiModel`. The TUI is then generated from the model via `View() string`.

Everything else is related to the `Reporter` interface which is a crude shim that injects outside events into the `Update` function.

build/devenv/reporter/summary.go
winder, line 1: This is the simple summary that will be displayed for non-TTY environments.

build/devenv/reporter/factory.go
```diff
@@ -0,0 +1,28 @@
...
+// terminal fd after stdout/stderr have been redirected to a log file).
+func New(verbose bool, out io.Writer) devenvruntime.Reporter {
+	if verbose {
+		return devenvruntime.NoopReporter{}
+	}
+	if f, ok := out.(*os.File); ok && term.IsTerminal(f.Fd()) {
+		return newBubbletearReporter(out)
+	}
+	return newSimpleReporter(out)
+}
```
winder, line 28: Extra fancy term stuff.

PR #1184, opened 2026-06-19
verifier: committee verifier multi-aggregator support

makramkd:
**Note to reviewers**: the diff looks large but almost 50% of it is test coverage. Unless you're specifically interested in the tests introduced (outlined below in the "Testing" section) its recommended to mark those as "Viewed" in the Files changed view to reduce the code to look at.

## Description

Today, the committee verifier achieves write high-availability by deploying one verifier job per aggregator. Each job carries a single `AggregatorAddress` and the deployment changeset emits N jobs (one per `committee.Aggregators` entry). This is operationally heavy: N job specs (1 per aggregator in the HA setup), more RPC endpoint load (since the same verifier logic is running redundantly), and this has become a source of inconsistency across deployment modes (CL node vs. standalone). The CL node supports multiple jobs, whereas standalone currently supports only one, for reasons of simplicity.

This PR lets a single consolidated verifier job write to multiple aggregators, so operators can collapse those N jobs into one. This also means that standalone and CL mode can achieve identical behavior as far as aggregator HA goes w/ the same committee verifier config.

What changes:

- **Config (backwards compatible).** `commit.Config` gains an `aggregators` list (`[[aggregators]]` array-of-tables, each with address / insecure flag / optional name + message-size limits). The legacy `aggregator_address` is retained: the list wins when set, otherwise a single entry is synthesized from the legacy fields, and setting both is a validation error. Existing production configs are unaffected.
- **Runtime fan-out.** Writes go to all aggregators with **all-must-ack** semantics. An item's durable job completes only when every aggregator acks; retries re-send to all (this is safe because aggregator writes are idempotent). Each aggregator gets its own resilient (circuit-breaker/timeout) + observed stack so one slow/down aggregator can't trip the others. Heartbeats fan out to every aggregator (single-aggregator failure is non-blocking). Disablement rules use a **fail-safe union**, i.e. a message is disabled if any aggregator disables it, and verification is blocked while any source's rules are unknown.
- **Deployment (opt-in, default off).** `ApplyVerifierConfig` and `AddNOPOffchain` gain a `ConsolidateAggregators` flag. Default `false` preserves byte-identical one-job-per-aggregator output. When `true`, one job per NOP is emitted using the `aggregators` list and an aggregator-name-free `verifier_id` (`NewConsolidatedVerifierJobID`). The consolidated job ID still matches `VerifierJobScope`, so flipping the flag lets `RevokeOrphanedJobs` clean up the old per-aggregator jobs.

Because the changeset flag defaults off and the config is additive, runtime + config + changeset land together with zero behavior change. Therefore, this new code can be rolled out w/out any config changes in production and should work out of the box.

Unknowns: 

1. It's unclear whether the changeset that was changed in this PR is the only one that is used across all families. It is probably used by Canton, but EVM might be using a different one.

## Testing

Unit tests added and run for every new decision point:

- `go test ./verifier/pkg/commit/`: config precedence (list wins / error if both / synthesize from legacy), duplicate/empty-address rejection, and array-of-tables round-trip under BurntSushi and pelletier.
- `go test ./integration/storageaccess/`: `FanOutWriter.merge` exhaustively table-tested across all per-aggregator outcome combinations (N=1,2,3: success / retryable / non-retryable), heterogeneous batches, error aggregation, and the short-result-slice synthesis path.
- `go test ./integration/pkg/heartbeatclient/`: partial failure is non-blocking; all-fail surfaces a joined error.
- `go test ./integration/pkg/messagerules/`: union semantics (disabled-wins, strict fail-safe on unknown).
- `go test ./deployment/changesets/ ./deployment/shared/`: legacy path emits one job per aggregator (unchanged `verifier_id`/`AggregatorAddress`); consolidated path emits one job per NOP with the aggregators list and an aggregator-name-free `verifier_id`; scope matching detects both.

**NOTE**: the devenv has NOT been updated to use the new config values - waiting on some feedback before diving in and doing that.

## Checklist

### DIFF
--- a/integration/pkg/constructors/committee_verifier.go
+++ b/integration/pkg/constructors/committee_verifier.go
@@ -28,9 +28,14 @@ import (
 	"github.com/smartcontractkit/chainlink-evm/pkg/chains/legacyevm"
 )
 
-// NewVerificationCoordinator starts the Committee Verifier with evm chains.
-// Signing is passed in because it's managed differently in the CL node vs standalone modes.
-// The DataSource is passed in from CL node, or created by standalone mode.
+// NewVerificationCoordinator builds a verifier coordinator. aggregatorSecret is the HMAC
+// credential used for every configured aggregator.
+//
+// TODO(CCIP-11776 follow-up): for a consolidated job each aggregator authenticates with its own
+// credential, so this should take a per-aggregator map keyed by AggregatorConnection.SecretName.
+// That is a breaking change to the Chainlink-node caller (ccvcommitteeverifier/delegate.go) and
+// lands together with that migration; until then this constructor applies one credential to all
+// aggregators, which is correct for the single-aggregator (legacy) configs in use today.
 func NewVerificationCoordinator(
 	lggr logger.Logger,
 	cfg commit.Config,
@@ -150,20 +155,23 @@ func NewVerificationCoordinator(
 		return nil, fmt.Errorf("invalid aggregator configuration: %w", err)
 	}
 
+	// The single provided credential is applied to every aggregator (see the constructor's TODO).
 	writeTargets := make([]storageaccess.AggregatorTarget, len(resolvedAggregators))
 	heartbeatTargets := make([]heartbeatclient.AggregatorTarget, len(resolvedAggregators))
 	for i, a := range resolvedAggregators {
 		writeTargets[i] = storageaccess.AggregatorTarget{
-			Label:               a.Name,
+			Label:               a.Label(),
 			Address:             a.Address,
 			Insecure:            a.InsecureConnection,
+			HMACConfig:          aggregatorSecret,
 			MaxSendMsgSizeBytes: a.MaxSendMsgSizeBytes,
 			MaxRecvMsgSizeBytes: a.MaxRecvMsgSizeBytes,
 		}
 		heartbeatTargets[i] = heartbeatclient.AggregatorTarget{
-			Label:    a.Name,
-			Address:  a.Address,
-			Insecure: a.InsecureConnection,
+			Label:      a.Label(),
+			Address:    a.Address,
+			Insecure:   a.InsecureConnection,
+			HMACConfig: aggregatorSecret,
 		}
 	}
 
@@ -173,7 +181,6 @@ func NewVerificationCoordinator(
 		writeTargets,
 		cfg.VerifierID,
 		lggr,
-		aggregatorSecret,
 		verifierMonitoring,
 	)
 	if err != nil {
@@ -212,7 +219,6 @@ func NewVerificationCoordinator(
 		heartbeatTargets,
 		cfg.VerifierID,
 		lggr,
-		aggregatorSecret,
 		heartbeat.NewHeartbeatMonitoringAdapter(verifierMonitoring),
 	)
 	if err != nil {
@@ -231,7 +237,7 @@ func NewVerificationCoordinator(
 
 	namedPollers := make([]messagerules.NamedPoller, 0, len(resolvedAggregators))
 	for _, a := range resolvedAggregators {
-		aggLggr := logger.With(lggr, "component", "MessageRulesPoller", "aggregator", a.Name)
+		aggLggr := logger.With(lggr, "component", "MessageRulesPoller", "aggregator", a.Label())
 		messageRulesClient, rErr := messagerules.NewGRPCClient(
 			a.Address,
 			aggLggr,
@@ -240,8 +246,8 @@ func NewVerificationCoordinator(
 			a.MaxRecvMsgSizeBytes,
 		)
 		if rErr != nil {
-			lggr.Errorw("Failed to create message rules gRPC client", "error", rErr, "aggregator", a.Name)
-			return nil, fmt.Errorf("failed to create message rules client for %q: %w", a.Name, rErr)
+			lggr.Errorw("Failed to create message rules gRPC client", "error", rErr, "aggregator", a.Label())
+			return nil, fmt.Errorf("failed to create message rules client for %q: %w", a.Label(), rErr)
 		}
 
 		poller, rErr := messagerules.NewPollerService(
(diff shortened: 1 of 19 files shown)


verifier/pkg/commit/config.go
```diff
@@ -12,20 +12,62 @@ import (
 // DefaultECDSASigningKeyName is the keystore key name for the ECDSA key used to sign verification results.
 const DefaultECDSASigningKeyName = "bootstrap_default_ecdsa_signing_key"
 
+// AggregatorConnection describes a single aggregator the verifier writes to, sends
+// heartbeats to, and reads message-disablement rules from. A consolidated verifier job
+// carries one of these per aggregator in Config.Aggregators.
+type AggregatorConnection struct {
```
makramkd, line 20: IMO we should have had the aggregator connection config like this from the start in the verifier config, for better overall organization and readability. This is basically all the existing fields with the addition of `Name` to be able to disambiguate b/w multiple aggregators.

verifier/pkg/commit/config.go
```diff
@@ -78,7 +120,71 @@ func (c *Config) MessageDisablementRulesClientTimeoutDuration() (time.Duration,
...
+	}
+
+	return resolved, nil
+}
+
 func (c *Config) Validate() error {
+	// Only validate aggregator wiring when something is configured. A config with neither
+	// aggregator_address nor aggregators set has always passed Validate (the gRPC client is
+	// created lazily), so we keep that lenient behavior; the strict "at least one aggregator"
+	// requirement is enforced by ResolvedAggregators on the startup path.
```
makramkd, line 284: This seems like an existing bug, we never validated that the aggregator_address was ever set; I'm inheriting this behavior but we should figure out whether to make this more strict.

integration/storageaccess/observed_storage.go
```diff
@@ -33,6 +37,35 @@ func NewObservedStorageWriter(
...
+	}
+}
+
+// metrics returns the metric labeler scoped to this observer's verifier_id and (when set)
+// aggregator label.
+func (o *observedStorageWriter) metrics() verifier.MetricLabeler {
+	m := o.monitoring.Metrics().With("verifier_id", o.verifierID)
+	if o.aggregatorLabel != "" {
+		m = m.With("aggregator", o.aggregatorLabel)
+	}
```
makramkd, line 65: This is purely additive - but I'm not sure if it would cause issues w/ our existing alerts cc @bukata-sa @emate .

integration/storageaccess/fanout_writer.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	return synthesized
+}
+
+// merge collapses the per-aggregator results into one result per item using all-must-ack
+// semantics:
+//   - success only when every aggregator acked the item;
+//   - on failure, retryable unless any aggregator returned a non-retryable error (retrying
+//     cannot help that aggregator, so the item is failed permanently and logged distinctly).
+func (f *FanOutWriter) merge(ccvDataList []protocol.VerifierNodeResult, perAggregator [][]protocol.WriteResult) []protocol.WriteResult {
```
makramkd, line 164: The main logic of the fanout is here - the rules I settled on when calling multiple aggregators is that we would continue retrying until all of them return a successful response. My reasoning here is that if we _don't_ do that then we are not behaving the same as the existing setup w/ 2 jobs, one for each aggregator, since they would separately retry until a success.

integration/storageaccess/fanout_writer.go
```diff
@@ -0,0 +1,231 @@
...
+			// Permanent partial write: some aggregators may have the item, but at least one
+			// rejected it non-retryably and re-sending will not help. The item will not be
+			// retried, so those aggregators will not receive it.
+			f.lggr.Errorw("Permanent partial write: item rejected non-retryably by an aggregator and will not be retried",
+				"messageID", data.MessageID.String(),
+				"sourceChain", data.Message.SourceChainSelector,
+				"failedAggregators", failedLabels,
+				"error", merged[i].Error,
+			)
+		}
```
makramkd, line 210: I'm not sure under which scenario this would occur in the existing setup w/ 2 jobs; in any case, this is equivalent to the existing setup because if job A gets nonRetryable and job B succeeds, both would move on from that message.

integration/pkg/messagerules/union_poller.go
```diff
@@ -0,0 +1,115 @@
...
+// fail-safe union semantics:
+//
+//   - A message is disabled if ANY aggregator's rules disable it (so no kill-switch set on any
+//     aggregator is silently ignored). Disabled takes precedence over unknown: a known
+//     disablement results in (true, nil) so the task is dropped rather than retried.
+//   - If no aggregator disables the message but ANY aggregator is in the unknown state (no
+//     successful poll yet), the result is (true, ErrMessageRulesStateUnknown) so verification is
+//     blocked and retried (strict fail-safe). One unreachable rules endpoint therefore blocks the
+//     whole consolidated job.
+type UnionPollerService struct {
```
makramkd, line 32: In the existing setup, if a message is disabled on aggregator-1 but not disabled on aggregator-2, the job that writes to aggregator-2 would still verify and send the message, which leads to inconsistent state. I'm not sure under which conditions this can happen, but I believe that it is a possibility.

With the union setup here, if either aggregator-1 OR aggregator-2 have a message disabled, the message will be skipped. This seems the safer option in the case of e.g. some kind of misconfig/wrong rollout of the message disablement stuff.

carte7000 (dismissed):

KodeyThomas (dismissed):

integration/pkg/constructors/committee_verifier.go
```diff
@@ -28,9 +28,14 @@ import (
...
-// Signing is passed in because it's managed differently in the CL node vs standalone modes.
-// The DataSource is passed in from CL node, or created by standalone mode.
+// NewVerificationCoordinator builds a verifier coordinator. aggregatorSecret is the HMAC
+// credential used for every configured aggregator.
+//
+// TODO(CCIP-11776 follow-up): for a consolidated job each aggregator authenticates with its own
+// credential, so this should take a per-aggregator map keyed by AggregatorConnection.SecretName.
+// That is a breaking change to the Chainlink-node caller (ccvcommitteeverifier/delegate.go) and
+// lands together with that migration; until then this constructor applies one credential to all
+// aggregators, which is correct for the single-aggregator (legacy) configs in use today.
```
makramkd, line 38: This will be addressed in a follow-up.

verifier/pkg/commit/config.go
```diff
@@ -2,30 +2,158 @@ package commit
...
+	Name string `toml:"name"`
+	// SecretName is the credential lookup key for this aggregator — the one join key used in both
+	// deployment modes to resolve the aggregator's HMAC credential (kept distinct from Name so the
+	// display label can change without re-wiring secrets):
+	//   - standalone: the env var pair VERIFIER_AGGREGATOR_<SECRETNAME>_API_KEY / _SECRET_KEY;
+	//   - Chainlink node: the secrets.toml entry whose VerifierID equals this value, which also
+	//     becomes the key in the per-aggregator credential map passed to the coordinator.
+	// Empty (the legacy single-aggregator path) falls back to the default un-suffixed credential
+	// variables. Secrets themselves never live in config — only this reference does.
+	SecretName string `toml:"secret_name"`
```
makramkd, line 32: Added to simplify secrets management - this is never the actual secret, just the name of the secret in either secrets.toml or the env var key.

carte7000 (approved):

faisal-chainlink (approved):

PR #1203, opened 2026-06-26
common/jd: replacement fallback on StartJob failure

makramkd:
## Description

This PR fixes a stall bug in the JD lifecycle manager where a failed replacement proposal left the verifier with no running job and no recovery path.

When a replacement proposal arrived, the manager stopped the old job before calling `StartJob` for the new one. If `StartJob` failed, the old job was gone. The manager entered `WaitingForJob` with nothing persisted to restart from. JD wouldn't re-push since the proposal was already in-flight, so the verifier would stall until an operator manually re-proposed.

This builds on the two-phase persistence introduced in #1201  (the `approved` row is now preserved by `SaveJob` during a replacement). The fix adds an explicit rollback path:

1. `SaveJob` only deletes the `pending` row — the old `approved` row survives until the new job is confirmed running.
2. On `StartJob` failure during a replacement, `DeletePendingJob` removes the new pending row, then the old job is restarted from the in-memory snapshot.
3. If the old job restart also fails, state transitions to `WaitingForJob` (the old `approved` row is still in the DB, so restart recovery still works on the next process restart).
4. A `UNIQUE(status)` constraint (migration 0003) prevents accumulation of pending rows from repeated failed replacements.

The README.md has been updated to reflect the new control flow introduced in #1201 and this PR.

## Testing

- Ran `go test ./common/jd/... -v -count=1`
- New unit tests covering the replacement failure paths:
  - `TestManager_EventLoop_Proposal_Replacement_StartFails_FallsBackToOldJob`: replacement `StartJob` fails; old job restarted; state stays `Running`.
  - `TestManager_EventLoop_Proposal_Replacement_StartFails_OldJobRestartAlsoFails`: both new and old job fail; state transitions to `WaitingForJob`.
- New store tests covering the two-row invariants:
  - `TestPostgresStore_SavePreservesApprovedRow`
  - `TestPostgresStore_LoadJob_PrefersApproved`
  - `TestPostgresStore_MarkJobApproved_WithTwoRows`
  - `TestPostgresStore_DeletePendingJob_*` (3 cases)

## Checklist

### DIFF
--- a/common/jd/store/pg.go
+++ b/common/jd/store/pg.go
@@ -67,26 +67,32 @@ func (s *PostgresStore) SavePendingJob(ctx context.Context, proposalID string, v
 
 // AcceptPendingJob promotes the pending record to approved, replacing any old approved record.
 // Returns true if a pending record was promoted, false if none existed.
-// Two separate statements are required: PostgreSQL data-modifying CTEs use the same snapshot
-// for both the CTE and the main query, so a combined DELETE+UPDATE CTE would violate the
-// UNIQUE(status) constraint even though the net result would be valid.
+// The DELETE and UPDATE run in a transaction so that a failure on the UPDATE rolls back the
+// DELETE — preventing the approved row from being silently lost.
+// A data-modifying CTE cannot be used instead because PostgreSQL CTEs share the same snapshot,
+// causing the UPDATE to see the approved row as still present and violate UNIQUE(status).
 func (s *PostgresStore) AcceptPendingJob(ctx context.Context) (bool, error) {
-	_, err := s.ds.ExecContext(ctx, `DELETE FROM job_store WHERE status = 'approved'`)
-	if err != nil {
-		return false, fmt.Errorf("failed to remove old approved job: %w", err)
-	}
+	var promoted bool
+	err := sqlutil.TransactDataSource(ctx, s.ds, nil, func(tx sqlutil.DataSource) error {
+		_, err := tx.ExecContext(ctx, `DELETE FROM job_store WHERE status = 'approved'`)
+		if err != nil {
+			return fmt.Errorf("failed to remove old approved job: %w", err)
+		}
 
-	result, err := s.ds.ExecContext(ctx,
-		`UPDATE job_store SET status = 'approved', updated_at = NOW() WHERE status = 'pending'`,
-	)
-	if err != nil {
-		return false, fmt.Errorf("failed to accept pending job: %w", err)
-	}
-	rows, err := result.RowsAffected()
-	if err != nil {
-		return false, fmt.Errorf("failed to get rows affected: %w", err)
-	}
-	return rows > 0, nil
+		result, err := tx.ExecContext(ctx,
+			`UPDATE job_store SET status = 'approved', updated_at = NOW() WHERE status = 'pending'`,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to accept pending job: %w", err)
+		}
+		rows, err := result.RowsAffected()
+		if err != nil {
+			return fmt.Errorf("failed to get rows affected: %w", err)
+		}
+		promoted = rows > 0
+		return nil
+	})
+	return promoted, err
 }
 
 // LoadJob retrieves the active job from the store.
--- a/common/jd/store/pg.go
+++ b/common/jd/store/pg.go
@@ -44,47 +44,44 @@ func NewPostgresStore(ds sqlutil.DataSource) *PostgresStore {
 	return &PostgresStore{ds: ds}
 }
 
-// SaveJob persists a new proposal as pending.
-// Any existing pending row is replaced; any existing approved row is preserved so that
-// a failed replacement can fall back to the old job on restart.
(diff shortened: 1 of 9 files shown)


common/jd/store/pg.go
bukata-sa: let's do a single db call
postgresql can do this
```
INSERT INTO job_store (
    proposal_id,
    version,
    spec,
    status,
    created_at,
    updated_at
)
VALUES ($1, $2, $3, 'pending', NOW(), NOW())
ON CONFLICT (status)
DO UPDATE
SET
    proposal_id = EXCLUDED.proposal_id,
    version = EXCLUDED.version,
    spec = EXCLUDED.spec,
    updated_at = NOW();
```

common/jd/store/pg.go
```diff
@@ -36,20 +36,23 @@ type PostgresStore struct {
...
 func NewPostgresStore(ds sqlutil.DataSource) *PostgresStore {
 	return &PostgresStore{ds: ds}
 }
 
-// SaveJob persists a job spec as pending, replacing any existing job.
-// Only one job should be active at a time.
+// SaveJob persists a new proposal as pending.
+// Any existing pending row is replaced; any existing approved row is preserved so that
+// a failed replacement can fall back to the old job on restart.
 func (s *PostgresStore) SaveJob(ctx context.Context, proposalID string, version int64, spec string) error {
```
bukata-sa: let's call the method `SavePendingJob`

common/jd/store/pg.go
```diff
@@ -64,24 +67,37 @@ func (s *PostgresStore) SaveJob(ctx context.Context, proposalID string, version
...
-		`UPDATE job_store SET status = 'approved', updated_at = NOW()`,
+	// Remove the old approved row first so the unique(status) constraint allows the update.
+	// A crash between the two statements leaves only a pending row → pending recovery path on restart.
+	_, err := s.ds.ExecContext(ctx, `DELETE FROM job_store WHERE status = 'approved'`)
+	if err != nil {
+		return fmt.Errorf("failed to remove old approved job: %w", err)
+	}
+
+	_, err = s.ds.ExecContext(ctx,
+		`UPDATE job_store SET status = 'approved', updated_at = NOW() WHERE status = 'pending'`,
```
bukata-sa: there is no such thing as UPDATE ... ON CONFLICT, but it is possible to have single query too
```
WITH deleted AS (
    DELETE FROM job_store
    WHERE status = 'approved'
)
UPDATE job_store
SET
    status = 'approved',
    updated_at = NOW()
WHERE status = 'pending';
```

common/jd/store/pg.go
```diff
@@ -64,24 +67,37 @@ func (s *PostgresStore) SaveJob(ctx context.Context, proposalID string, version
 	return nil
 }
 
-// MarkJobApproved transitions the stored job's status from pending to approved.
+// MarkJobApproved promotes the pending record to approved, atomically replacing the old
+// approved record (if any). Called after StartJob succeeds during a replacement.
 func (s *PostgresStore) MarkJobApproved(ctx context.Context) error {
```
bukata-sa: I would also call that smth like `AcceptPendingJob`
I intuitively wait a job argument in `MarkJobApproved` method

common/jd/store/pg.go
```diff
@@ -64,24 +67,37 @@ func (s *PostgresStore) SaveJob(ctx context.Context, proposalID string, version
...
+// When both an approved and a pending row exist (failed replacement), the approved row
+// is returned so the known-good job starts on restart.
 // Returns ErrNoJob if no job is found.
 func (s *PostgresStore) LoadJob(ctx context.Context) (*Job, error) {
 	var rows []jobRow
 	err := s.ds.SelectContext(ctx, &rows,
 		`SELECT proposal_id, version, spec, status, created_at, updated_at
-		 FROM job_store ORDER BY id DESC LIMIT 1`,
+		 FROM job_store
+		 ORDER BY CASE WHEN status = 'approved' THEN 0 ELSE 1 END, id DESC
```
bukata-sa: nit: this can be `ORDER BY (status = 'approved') DESC, id DESC`
but if it is less obvoius, we can keep as it is

common/jd/store/pg.go
bukata-sa: Also I think that this method needs a feedback was there a pending job to approve or not
This sql returns amount of rows updated so you can `return rows > 0, nil` to show how if there were a job to approve

common/jd/store/pg.go
```diff
@@ -117,12 +133,22 @@ func (s *PostgresStore) HasJob(ctx context.Context) (bool, error) {
 	return counts[0] > 0, nil
 }
 
-// DeleteJob removes the persisted job from the store.
-// This is called when JD sends a delete request.
+// DeleteJob removes all persisted job records.
+// Called when JD sends a delete request.
 func (s *PostgresStore) DeleteJob(ctx context.Context) error {
```
bukata-sa: Maybe `DeleteAllJobs`? Or just `Clear` since this is a job store

bukata-sa (dismissed):

carte7000 (approved):

bukata-sa (approved):

PR #1205, opened 2026-06-26
bootstrap: inform JD of chain support and public keys

makramkd:
## Description

Deployment changesets (e.g. `apply_verifier_config`) need each committee verifier node's onchain signing address before they can generate a job spec or configure the onchain contract. Prior to this PR, standalone verifiers had no mechanism to publish their signing key to JD. devenv worked around this by reading keys directly from the bootstrap HTTP server and injecting them into the topology at deploy time, but production deployments had no equivalent path.

This PR makes standalone verifier nodes push their signing key to JD on connect via `feedsmanager.UpdateNode`, matching how the Chainlink node publishes OCR2 key bundles. The address is stored in `OCR2Config.OcrKeyBundle.OnchainSigningAddress`. This is a deliberate shoehorn into the existing OCR2 chain config shape that `fetch_signing_keys.go` already reads. 

**What changed:**
- Bootstrap config gains an optional `[[chains]]` section (`type` + `id`). On JD connect the bootstrapper reads each `ECDSA_S256` key declared via `WithKey`, derives the chain-family-specific signing address, and calls `feedsmanager.UpdateNode` with one `ChainConfig` per declared chain. Supported families: EVM (EIP-55 checksummed), Solana (20-byte keccak256, no 0x), Aptos (raw uncompressed pubkey hex). Stellar and Canton are blocked on JD proto additions.
- `lifecycle.Manager` gains an optional `OnConnectHook` called after each JD connect. Failures are logged as warnings and do not block job processing.
- Devenv populates `[[chains]]` from `blockchain.Output.ChainID` at launch time and stops pre-filling `SignerAddressByFamily` from `BootstrapKeys` for families that now push via JD, letting the changeset's existing JD fallback run instead.

## Example Config

The following can be added to the bootstrap config TOML file, which is provided by the standalone node operator:

```toml
[[chains]]
  type = "evm"
  id = "1337"

[[chains]]
  type = "evm"
  id = "2337"

[[chains]]
  type = "evm"
  id = "3337"
```

On startup, upon connecting to JD, the node will send the appropriate `ChainConfig`s to JD w/ the signing keys for that family.

## Testing

- `go test ./bootstrap/... ./common/jd/... -count=1 -short`.
- `just lint fix`.
- Local devenv spinup + test; should work as expected while fetching standalone keys from JD instead of the manual path.

## Checklist


bootstrap/bootstrap.go
```diff
@@ -238,6 +241,105 @@ func (b *Bootstrapper) startWithAppConfig(ctx context.Context) (startErr error)
...
+
+	resp, err := ks.GetKeys(ctx, keystore.GetKeysRequest{KeyNames: signingKeyNames})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get signing keys from keystore: %w", err)
+	}
+	if len(resp.Keys) == 0 {
+		return nil, fmt.Errorf("no signing keys found in keystore for names %v", signingKeyNames)
+	}
+	// The committee verifier uses exactly one ECDSA_S256 signing key.
+	signingKey := resp.Keys[0]
```
makramkd: **nit**: the comment refers to ECDSA_S256 signing keys but in theory the verifier can use whichever type of key.

bootstrap/bootstrap.go
```diff
@@ -262,11 +364,37 @@ func (b *Bootstrapper) startWithJDLifecycle(ctx context.Context) error {
 	}
 
 	jobRunner := &runner{fac: b.fac, deps: deps}
+
+	var signingKeyNames []string
+	for _, k := range b.keys {
+		if k.keyType == keystore.ECDSA_S256 {
+			signingKeyNames = append(signingKeyNames, k.name)
+		}
+	}
```
makramkd, line 388: **question**: how do we know that these are the keys that were specified via the `WithKey` option?

carte7000 (dismissed):

bootstrap/bootstrap.go
```diff
@@ -265,6 +268,104 @@ func (b *Bootstrapper) startWithAppConfig(ctx context.Context) (startErr error)
 	return b.fac.Start(ctx, js, ServiceDeps{Registry: b.accCloser})
 }
 
+// chainTypeFromString maps a config chain type string to the proto ChainType enum.
+func chainTypeFromString(s string) (pb.ChainType, error) {
```
KodeyThomas, line 274: nit: are we able to eventually abstract this away as part of config? seems slightly painful to keep a chain specific mapping here

KodeyThomas (dismissed):
looks good, long term lets see if we can remove the chain types from bootstrap

bootstrap/bootstrap.go
makramkd: I had the same thought - the config input is currently a string like "evm" or "solana". In this particular code snippet we're limited by the JD protos, so maybe there's a "string to enum" conversion that is autogenerated w/ the pb files, will check.

makramkd: I think you can do this:

```golang
func chainTypeFromString(s string) (pb.ChainType, error) {
    key := "CHAIN_TYPE_" + strings.ToUpper(s)
    if v, ok := pb.ChainType_value[key]; ok {
        return pb.ChainType(v), nil
    }
    return pb.ChainType_CHAIN_TYPE_UNSPECIFIED, fmt.Errorf("unknown chain type %q", s)
}
```

will test out and report back.

makramkd: Yup, this works, will switch to it. We'd still need to bump the pb, but that's normal.

carte7000 (approved):

tt-cll (approved):

PR #1207, opened 2026-06-26
Standalone mode improvement step 2: deprecate blockchain info from appConfig for verifier serviceFactory

huangzhen1997:
Stop enumerating verifier source chains from job blockchain_infos. Use on_ramp_addresses instead, via the same chainaccess.Infos.GetAllChainSelectors() helper (non-numeric keys are skipped, matching the old behavior).

blockchain_infos remains in the app config for EVM accessor RPC via chainaccess.NewRegistry; Solana RPC still comes from the NOP file. Executor is unchanged, it already uses chain_configuration.

Related [doc](https://docs.google.com/document/d/1LGA_tvAqDrKlgTk9q2yzlNSvYm7jSZx2iLdqQT5TN3w/edit?tab=t.0): step 2

### DIFF
--- a/bootstrap/job.go
+++ b/bootstrap/job.go
@@ -23,8 +23,8 @@ type JobSpec struct {
 // JD app config should not ship chain connection info (RPC URLs, etc.); that belongs
 // in local config for standalone mode or node config for CL mode. Devenv still
 // injects blockchain_infos through this decode path today. Prefer GetAppConfig for
-// typed app-only config. GenericConfig should be fully deprecated
-func (js JobSpec) GetGenericConfig() (chainaccess.GenericConfig, error) {
+// typed app-only config. GenericConfig should be fully deprecated once EVM swtich to use local config
+func (js JobSpec) GetGenericConfig() (chainaccess.GenericConfig, error) { //nolint:staticcheck
 	var gcfg chainaccess.GenericConfig
 	if _, err := toml.Decode(js.AppConfig, &gcfg); err != nil {
 		return chainaccess.GenericConfig{}, fmt.Errorf("error decoding app config: %w", err)
--- a/bootstrap/job.go
+++ b/bootstrap/job.go
@@ -23,8 +23,8 @@ type JobSpec struct {
 // JD app config should not ship chain connection info (RPC URLs, etc.); that belongs
 // in local config for standalone mode or node config for CL mode. Devenv still
 // injects blockchain_infos through this decode path today. Prefer GetAppConfig for
-// typed app-only config. GenericConfig should be fully deprecated once EVM swtich to use local config
-func (js JobSpec) GetGenericConfig() (chainaccess.GenericConfig, error) { //nolint:staticcheck
+// typed app-only config. GenericConfig should be fully deprecated once EVM swtich to use local config.
+func (js JobSpec) GetGenericConfig() (chainaccess.GenericConfig, error) {
 	var gcfg chainaccess.GenericConfig
 	if _, err := toml.Decode(js.AppConfig, &gcfg); err != nil {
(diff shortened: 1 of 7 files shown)


bootstrap/job.go
```diff
@@ -35,6 +35,8 @@ func (js JobSpec) GetAppConfig(cfg any) error {
 		return fmt.Errorf("error decoding app config: %w", err)
 	}
 
+	// Deprecated: blockchain_infos should be dropped from appConfig, blockchain info will come from local config for standalone mode
+	// or node config for cl-mode. Remove this filter once chain family repos stopped using blockchain_infos in appConfig.
```
makramkd, line 45: **nit**: I think we should add the `// Deprecated` comment to probably `GetGenericConfig` as well, because the `GenericConfig` is the place where the `blockchain_infos` TOML struct tag is defined and we use it explicitly to ship things like RPC URLs and such in devenv. In follow-ups we should maybe fully deprecate `GenericConfig`, and the job spec only ever ships app config.

cmd/verifier/servicefactory.go
```diff
@@ -73,8 +71,6 @@ func (f *factory) Start(ctx context.Context, spec bootstrap.JobSpec, deps bootst
 		f.profiler = profiler
 	}
 
-	chainSelectors := genericConfig.ChainConfig.GetAllChainSelectors()
-
```
makramkd, line 77: **question**: do we need to do something similar for the executor factory?

huangzhen1997: No, executor uses getAppConfig() and loop over [executorConfig.ChainConfiguration](https://github.com/smartcontractkit/chainlink-ccv/blob/351c264b596a6d0c5386555c118b519583a2e7b3/cmd/executor/service.go#L103).  Not getGenericConfig() with blockchain_info for enumeration. So it's  fine.

huangzhen1997: There is a blockchain_info field [here](https://github.com/smartcontractkit/chainlink-ccv/blob/351c264b596a6d0c5386555c118b519583a2e7b3/executor/config.go#L26) which we should remove, but the numeration uses [ChainConfiguration](https://github.com/smartcontractkit/chainlink-ccv/blob/351c264b596a6d0c5386555c118b519583a2e7b3/executor/config.go#L65)

integration/pkg/accessors/evm/factory_constructor.go
```diff
@@ -37,6 +37,7 @@ var _ chainaccess.AccessorFactoryConstructor = CreateEVMAccessorFactory
 func CreateEVMAccessorFactory(lggr logger.Logger, genericConfig chainaccess.GenericConfig) (chainaccess.AccessorFactory, error) {
 	// Convert generic chain config -> Infos[evm.Info]
 	evmInfos := make(chainaccess.Infos[Info])
+	// TODO: To support standalone mode, need to support local config, GenericConfig will be deprecated as JD job will not have rpc info.
```
huangzhen1997, line 40: @makramkd Is there a potential future support for EVM standalone mode? It wouldn't work now since there's no local config, it only works for devenv with that manual rebuild patch to insert ctf rpc nodes into blockchain_infos.

We can't remove genericConfig now as EVM is the one still using it.

makramkd: Yeah, we can't remove it yet, I can create a follow up to switch EVM to use a mounted config.

makramkd (dismissed):

integration/pkg/accessors/evm/factory_constructor.go
makramkd: [Ticket](https://smartcontract-it.atlassian.net/browse/CCIP-11910) for EVM.

makramkd (approved):

jadepark-dev (approved):

PR #1209, opened 2026-06-29
feat: parallelise attestation fetching and verification

kirtan-amin:
## Description

We currently fetch and verify attestations in a synchronous manner, these changes parallelises this by using fixed set of workers in a pool.

## Checklist

### DIFF
--- a/verifier/pkg/token/cctp/verifier.go
+++ b/verifier/pkg/token/cctp/verifier.go
@@ -16,6 +16,8 @@ import (
 const (
 	attestationNotReadyRetry = 30 * time.Second
 	anyErrorRetry            = 5 * time.Second
+	// Max number of concurrent workers to fetch attestations to verify.
+	maxAttestationFetchers = 10
 )
 
 // Verifier is responsible for verifying CCTP messages by fetching their attestations
@@ -27,6 +29,7 @@ type Verifier struct {
 
 	attestationNotReadyRetry time.Duration
 	anyErrorRetry            time.Duration
+	maxAttestationFetchers   int
 }
 
 func NewVerifier(
@@ -38,6 +41,7 @@ func NewVerifier(
 		attestationService,
 		attestationNotReadyRetry,
 		anyErrorRetry,
+		maxAttestationFetchers,
 	)
 }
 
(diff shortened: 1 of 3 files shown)


verifier/pkg/token/cctp/verifier.go
```diff
@@ -46,84 +50,100 @@ func NewVerifierWithConfig(
...
-
-		lggr.Debugw("Attestation fetched and decoded successfully",
-			"status", attestation.status,
-			"attestation", attestation.attestation,
-			"encodedCCTPMessage", attestation.encodedCCTPMessage,
-			"verifierFormat", verifierFormat,
-		)
+		jobs <- task
+	}
+	close(jobs)
```
KodeyThomas:
```suggestion
    defer close(jobs)
```

`defer` should make this run after the function has returned. should save some issues reading from a closed channel (which causes a panic)

KodeyThomas (changes requested):

verifier/pkg/token/cctp/verifier.go
```diff
@@ -16,6 +16,8 @@ import (
 const (
 	attestationNotReadyRetry = 30 * time.Second
 	anyErrorRetry            = 5 * time.Second
+	// Max number of concurrent workers to fetch attestations to verify.
+	maxAttestationFetchers = 10
```
carte7000, line 20: nit: Maybe we should consider adding this to the token verifier config. Currently the verifiers(committee verifier included) have a lot of these type of hardcoded config that we can't really change between env

verifier/pkg/token/cctp/verifier.go
kirtan-amin: Hmm, thinking about this a bit more and not sure we need the defer. The jobs channel has items sent to it during the loop above it, so once it's done there will be no risk of sending to a closed channel (which is the panic risk AFAIK). 

lmk if my thinking is right here!

verifier/pkg/token/cctp/verifier.go
kirtan-amin: Makes sense, will roll this into a separate PR

KodeyThomas (approved):

carte7000 (approved):

verifier/pkg/token/cctp/verifier.go
kirtan-amin: Addressed in https://github.com/smartcontractkit/chainlink-ccv/pull/1213

PR #1214, opened 2026-06-30
CCIP-11795 CCIP-12199 log streaming in standalone verifier/executor

bukata-sa:
https://smartcontract-it.atlassian.net/browse/CCIP-11795
## Description
Setup beholder and logger in bootstrap against bootstrap config. In order for log streaming to work, at the moment of logger creation (or zap core, particularly), beholder should be initialized
Service factories are no longer setup beholder/logging but use ones provided by bootstrap

Service factories now require to define metric views. Service has no histogram or default are applicable, pass a `nil`

## Testing
```
just rebuild-all
```
Check that executors, verifiers and token-verifiers tries to call otel /v1/logs endpoint
```
just obs up
just test smoke-v3 --build false
```
check that logs are flowing in loki

## Checklist


common/monitoring/logging/logging.go
```diff
@@ -1,6 +1,41 @@
...
+	"go.uber.org/zap/zapcore"
+
+	"github.com/smartcontractkit/chainlink-ccv/pkg/monitoring"
+	zaplog "github.com/smartcontractkit/chainlink-ccv/protocol/common/logging"
+	"github.com/smartcontractkit/chainlink-common/pkg/beholder"
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/logger/otelzap"
+)
+
+func InitLogger(name string, baseLogLevel zapcore.Level, config monitoring.BeholderConfig) (logger.Logger, error) {
```
makramkd: **nit**: lets add a go doc comment for this common helper.

bootstrap/bootstrap.go
```diff
@@ -38,9 +37,6 @@ const (
 
 // ServiceDeps are the dependencies passed to the services started by the bootstrapper.
 type ServiceDeps struct {
-	// Logger is a logger that can be used by the service.
-	Logger logger.Logger
-
```
makramkd, line 46: **question**: is there any reason to delete this if we're going to add it back soon?

bootstrap/bootstrap.go
```diff
@@ -224,10 +199,18 @@ func NewBootstrapper(
...
-		lggr.Infow("loaded bootstrap config")
+	// do not fallback to it in bootstrapper
+	mon := monitoring.BeholderConfig{}
+	if b.config != nil && b.config.Monitoring != nil {
+		mon = b.config.Monitoring.Beholder
+	}
+	lggr, err := common.InitLogger(b.name, b.logLevel, mon)
+	if err != nil {
+		return nil, fmt.Errorf("failed to init logger: %w", err)
 	}
```
makramkd: **issue**: the bootstrap logger uses the user-provided level (via WithLogLevel) but this log level doesn't get passed down to the service factories right now; I think this'd be a surprising change for users of bootstrap (i.e. chain integrations).

bootstrap/bootstrap.go
```diff
@@ -163,12 +140,10 @@ type Bootstrapper struct {
 // NewBootstrapper creates a new [Bootstrapper] with the given config and service factory.
 func NewBootstrapper(
 	name string,
-	lggr logger.Logger,
 	fac ServiceFactory,
 	opts ...Option,
 ) (*Bootstrapper, error) {
```
makramkd, line 154: **note**: this is currently a public-facing API (though one could argue, maybe it shouldn't be). If we make a breaking change like this it should eventually be documented w/ a changelog (see the changelog/ directory for examples and the template for the AI).

bootstrap/bootstrap.go
```diff
@@ -38,9 +37,6 @@ const (
 
 // ServiceDeps are the dependencies passed to the services started by the bootstrapper.
 type ServiceDeps struct {
-	// Logger is a logger that can be used by the service.
-	Logger logger.Logger
```
makramkd, line 45: **question**: to preserve the app-specified log levels while also being backwards compatible, and also make it easier in the future to just use the beholder logger from the bootstrap:

```golang
type LoggerWithLevel interface {
  logger.Logger
  Level() zapcore.Level
}

type loggerWithLevel struct {
  logger.Logger
  level zapcore.Level
}

func (l *loggerWithLevel) Level() zapcore.Level { return l.level }

type ServiceDeps struct {
  // Logger is a logger that can be used by the service.
  Logger LoggerWithLevel

  // rest as before...
}

// in NewBootstrapper
b.lggr = &loggerWithLevel{lggr, b.logLevel}

// when creating ServiceDeps
deps := ServiceDeps{
  Logger: b.lggr, // added
  // rest as before...
}

// in e.g. verifier ServiceFactory
// TODO: use deps.Logger after making bootstrap config required
f.lggr, err = common.InitLogger("verifier", deps.Logger.Level(), config.Monitoring.Beholder)
```

Any `LoggerWithLevel` object can be assigned to a `logger.Logger` object, as its a interface superset.

common/logging.go
```diff
@@ -1,6 +1,41 @@
...
+			config.LogStreamingLevel = "info"
+		}
+		logStreamingLevel, err := zapcore.ParseLevel(config.LogStreamingLevel)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing streaming log level: %w", err)
+		}
+		otelCore := otelzap.NewCore(beholder.GetLogger(), otelzap.WithLevel(logStreamingLevel))
+		loggerCores = append(loggerCores, otelCore)
+	}
+	zap.NewAtomicLevel()
```
makramkd: **issue**: seems unused

cmd/verifier/servicefactory.go
```diff
@@ -62,7 +58,12 @@ func (f *factory) Start(ctx context.Context, spec bootstrap.JobSpec, deps bootst
 	if err := spec.GetAppConfig(&config); err != nil {
 		return fmt.Errorf("failed to get app config: %w", err)
 	}
-	lggr.Infow("Using blockchain information from config", "info", genericConfig.ChainConfig)
+
+	lggr := deps.Logger
```
makramkd: **question**: is there a reason we removed the `logger.Sugared(logger.Named(...))` part?

bootstrap/bootstrap.go
```diff
@@ -49,32 +49,6 @@ type ServiceDeps struct {
 
 	// Registry for chainaccess.Accessor objects.
 	Registry chainaccess.Registry
-
-	// Monitoring is the operator-provided monitoring config from the bootstrap config (Config.Monitoring).
-	// It is nil when the operator did not configure monitoring in the bootstrap config, or when the
-	// bootstrapper runs in static-TOML mode (which loads no bootstrap config). Services prefer this value
-	// and fall back to their own app-config monitoring field when it is nil.
-	Monitoring *monitoring.Config
```
makramkd, line 56: **question**: why remove the config that gets piped to the services? Don't they need to also setup their monitoring using the same config?

bukata-sa: it was needed only to init beholder
now it is done in bootstrap

cmd/verifier/servicefactory.go
bukata-sa: it makes no effect, we never used sugared interface

makramkd (approved):

KodeyThomas (approved):

PR #1216, opened 2026-06-30
verifier/token: consume bootstrap config toml

makramkd:
## Description

The token verifier previously sourced monitoring config from its own app config TOML, which was fine when monitoring was the only operator-provided section. The blocker was that `bootstrap.Config` unconditionally required the full infra bundle (JD/DB/Keystore/Server), so the token verifier, which runs in static-TOML mode with no DB or JD,  couldn't load a bootstrap config at all.

The trigger to fix this arrived: a structured OTel log-streaming logger will be constructed in bootstrap code from `monitoring.Config` and delivered to service factories via `ServiceDeps.Logger`. The token verifier must receive that logger, which means it must participate in the same operator-config load path as the commit verifier and executor.

**What changed:**

- bootstrap.Config.validate() is now mode-driven. It validates the infra bundle ([jd], [db], [keystore], [server]) only in JD mode, keyed on the jdMode flag (passed as needsInfra), not on which sections happen to be present in the TOML. 
  - In static-TOML mode the infra bundle is ignored: any infra section present is logged as a warning, not an error. Monitoring is always validated independently, in both modes. This makes a monitoring-only bootstrap TOML valid, while a JD-mode app with a missing/malformed infra section still fails at load time with a precise, per-section error.

- `Bootstrapper` gains an unexported `jdMode bool`. `Start()` routes on this flag instead of `b.config != nil`, because `b.config` can now be set in both modes. In static-TOML mode, if `BOOTSTRAPPER_CONFIG_PATH` is explicitly set, the bootstrap TOML is loaded (no default fallback to `/etc/config.toml`; that would collide with `TOKEN_VERIFIER_CONFIG_PATH`, which shares the same default; will create a follow-up ticket for this).

- `startWithAppConfig` now builds a full `ServiceDeps` (Logger + Monitoring). Previously it only set `Registry`. The token verifier's inline logger-creation block (and its `// TODO: use deps.lggr` comment) is gone; it uses `deps.Logger` via `WithLogLevelFromEnv`.

- The token verifier uses `bootstrap.ResolveMonitoring(deps.Monitoring, cfg.Monitoring)` — the same F2 fallback pattern as the commit verifier and executor — so existing deployments without a bootstrap config continue to read monitoring from the app config unchanged.

- Devenv generates a separate bootstrap config (monitoring only) for the token verifier alongside its app config, and sets both `TOKEN_VERIFIER_CONFIG_PATH` and `BOOTSTRAPPER_CONFIG_PATH` explicitly. This also fixes the pre-existing TODO where the app config was mounted at `aggregator.DefaultConfigFile`.

The change is backwards compatible. No operator action is required until they want to opt in; see the migration guide in `changelog/2026-07-01_token_verifier_operator_config.md`.

## Testing

- `go test ./bootstrap/...`.

## Checklist

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -216,7 +216,7 @@ func NewBootstrapper(
 
 		b.configPath = resolveBootstrapConfigPath(b.configPath)
 		b.config = &Config{}
-		if err := LoadAndValidateConfig(b.configPath, b.config); err != nil {
+		if err := LoadAndValidateConfig(lggr, b.configPath, b.config, true); err != nil {
 			return nil, fmt.Errorf("failed to load bootstrap config (%s): %w", b.configPath, err)
 		}
 		// not logging config because it contains secrets.
@@ -227,7 +227,7 @@ func NewBootstrapper(
 		// here: TOKEN_VERIFIER_CONFIG_PATH and BOOTSTRAPPER_CONFIG_PATH both default to
 		// /etc/config.toml, so applying the default would decode the wrong file. See issue #013.
 		b.config = &Config{}
-		if err := LoadAndValidateConfig(path, b.config); err != nil {
+		if err := LoadAndValidateConfig(lggr, path, b.config, false); err != nil {
 			return nil, fmt.Errorf("failed to load operator config (%s): %w", path, err)
 		}
 		lggr.Infow("loaded operator config for static-TOML mode")
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -210,46 +210,27 @@ func NewBootstrapper(
 	if b.jdMode {
 		// JD mode requires a CSA key for node authentication. Inject the default if the caller
 		// did not explicitly declare one, so callers only need to list their application keys.
-		hasCSA := false
-		for _, k := range b.keys {
-			if k.purpose == "csa" {
-				hasCSA = true
-				break
-			}
-		}
-		if !hasCSA {
+		if !hasCSAKey(b.keys) {
 			b.keys = append([]keyToInit{{DefaultCSAKeyName, "csa", keystore.Ed25519}}, b.keys...)
 		}
 
-		// Use provided config path if set by an option.
-		if b.configPath == "" {
-			// If no config path is provided by an option, check the environment variable.
-			b.configPath = os.Getenv(ConfigPathEnv)
-			if b.configPath == "" {
-				// If the environment variable is not set, use the default config path.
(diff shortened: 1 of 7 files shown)


cmd/verifier/token/main.go
```diff
@@ -96,27 +96,7 @@ func (tvf *tokenVerifierFactory) Start(ctx context.Context, spec bootstrap.JobSp
...
-		tvf.lggr, err = logger.NewWith(logging.GetLogProfile(zapLevel))
-		if err != nil {
-			return fmt.Errorf("failed to create logger: %v", err)
-		}
-		tvf.lggr = logger.Named(tvf.lggr, "verifier")
-	}
-
-	// Use SugaredLogger for better API
-	tvf.lggr = logger.Sugared(tvf.lggr)
+	tvf.lggr = deps.Logger
```
makramkd, line 99: **nit**: bring back Sugared?

bootstrap/config.go
```diff
@@ -143,7 +143,19 @@ type Config struct {
 	Monitoring *monitoring.Config
 }
 
-func (c *Config) validate() error {
+// hasInfraKeys reports whether any infra section (jd, db, keystore, server) appears in md.
+func hasInfraKeys(md toml.MetaData) bool {
+	for _, key := range md.Keys() {
+		switch key[0] {
```
makramkd: **nit**: this seems funky, should we use `key.String()` instead?

bootstrap/bootstrap.go
```diff
@@ -192,44 +197,40 @@ func NewBootstrapper(
...
+		// Static-TOML mode: optionally load operator config when BOOTSTRAPPER_CONFIG_PATH is
+		// explicitly set. The default fallback to DefaultConfigPath is intentionally suppressed
+		// here: TOKEN_VERIFIER_CONFIG_PATH and BOOTSTRAPPER_CONFIG_PATH both default to
+		// /etc/config.toml, so applying the default would decode the wrong file. See issue #013.
+		b.config = &Config{}
+		if err := LoadAndValidateConfig(lggr, path, b.config, false); err != nil {
+			return nil, fmt.Errorf("failed to load operator config (%s): %w", path, err)
+		}
+		lggr.Infow("loaded operator config for static-TOML mode")
 	}
```
bukata-sa, line 234: what if not jd and env is not set? maybe return an error?

bukata-sa (approved):

KodeyThomas (approved):

bukata-sa (approved):

faisal-chainlink (approved):

PR #1221, opened 2026-07-01
feat(devenv/tcapi): bound persistent-network event scans, add delivery-only Run mode

jadepark-dev:
## Description

Two independent devenv/tcapi changes needed to run e2e tests against persistent staging chains instead of only local devenv:

1. `build/devenv/evm`: the event poller's first `eth_getLogs` scan ran unbounded (`[1, latest]`), which blows up against a long-lived chain like Sepolia and exceeds RPC provider block-range limits. Bounds the first scan to the last 1500 blocks (`fallbackLookbackBlocks`, mirroring `verifier/pkg/sourcereader.DefaultMaxBlockRange`), since the target event is always recent when a waiter registers. Also makes onRamp/offRamp poller construction lazy, so a test only pays for the pollers it actually uses.

2. Presence-based offchain assertions. `ccv.Lib`'s plural getters `AllAggregators`/`AllIndexers` now return an empty result with a nil error when the environment declares no aggregator/indexer endpoints, instead of erroring; the singular getters keep their original error contract. A shared helper `tcapi.SetupOffchainClients` wires the aggregator/indexer clients when configured and returns nil clients when not, and `basic.Run`/`token_transfer.Run` gate their aggregator/indexer assertions on client presence (`AssertMessage` already skips stages for nil clients). 

## Testing

- Local (`env-phased-out.toml`): `TestSolana2EVM` and `TestEVM2Solana` (all subtests, including the 4 TokenTransfer pool combinations and ArbMessaging) pass; endpoints are configured, so offchain assertions run as before.
- Unit: table-driven tests for the `Lib` getter absence/error contract and for `SetupOffchainClients` (present / absent / broken).
- Staging (`env-private-staging.toml`, Sepolia <-> Solana devnet): the presence path correctly detects no offchain endpoints and runs on-chain-only

## Requires
- https://github.com/smartcontractkit/chainlink-ccv/pull/1223

## Checklist


build/devenv/evm/event_poller.go
```diff
@@ -11,6 +11,12 @@ import (
 	"github.com/smartcontractkit/chainlink-ccv/protocol"
 )
 
+// fallbackLookbackBlocks bounds how far back the first eth_getLogs scan reaches on a
+// long-lived chain, so it neither scans from genesis nor exceeds the provider's
+// getLogs range limit — in blocks, not wall-clock, since the target event is always
+// recent. Mirrors verifier/pkg/sourcereader's DefaultMaxBlockRange.
+const fallbackLookbackBlocks uint64 = 1500
```
jadepark-dev, line 18: Kept as an independent constant rather than importing `sourcereader.DefaultMaxBlockRange` to avoid a devenv>verifier package dependency

build/devenv/tests/e2e/tcapi/basic/v3.go
```diff
@@ -111,20 +111,15 @@ func (tc *v3TestCase) Run(ctx context.Context) error {
...
-	if tc.aggregatorQualifier != "" && tc.aggregatorQualifier != common.DefaultCommitteeVerifierQualifier {
-		if client, ok := aggregatorClients[tc.aggregatorQualifier]; ok {
-			aggregatorClient = client
+	var aggregatorClient *ccv.AggregatorClient
+	var indexerMonitor *ccv.IndexerMonitor
+	if !tc.args.Run.OnchainAssertionOnly {
+		var setupErr error
+		aggregatorClient, indexerMonitor, setupErr = setupAggregatorAndIndexer(tc.lib, tc.aggregatorQualifier)
+		if setupErr != nil {
+			return setupErr
```
makramkd: **question**: I had a similar idea to make aggregator/indexer assertions optional on environments where this information may not be available - was wondering if we should gate this purely based on presence (i.e. `len(aggregatorClients) == 0` means no aggregator assertions - or if `aggregatorClient, ok` returns `nil, false`, skip the assertion) vs. doing it via flag (OnchainAssertionOnly).

The main advantage I see to the presence-based approach is that you don't need to update the test code or write a new test in order to run it in different environments - but curious as to what your. thoughts are.

jadepark-dev: Valid point, `TestingContext.AssertMessage` seems to be presence-based already, but iirc there was hard configuration validation. Maybe I can split that logic into absence vs. connection failure on lib construction. Let me explore a bit

makramkd (approved):

agusaldasoro (approved):

PR #1226, opened 2026-07-02
feat: migrate chaos tests to e2e/chaos package and V3 RunScenario

Farber98:
### Why
Chaos e2e tests embedded `Pumba` commands, container lookup, and a V2 send/assert path directly in `chaos_test.go`. That was fine for EVM-only, but now we need chaos tests in other families and avoid copying all primitives and setup. (eg. `Pumba`, env-out parsing, etc.)

This PR extracts shared chaos helpers and migrates all `TestChaos_*` tests onto a single V3 scenario runner.

### What
**New `tcapi/chaos/` package:**
- Outage injection via `Pumba` (`InjectOutage`, duration constants)
- Container resolution from `ccv.Cfg` (aggregator nginx, verifiers, executors, indexer)
- `ExecutorContainersForDest` for dest-chain executor targeting (needed when multiple executors share the same qualifier)
- `RunScenario` — outage → V3 send → offchain assert → optional dest exec confirm
- `ResolveEOAReceiverDefaultVerifier` for EVM chaos message setup

**chaos_test.go refactor:**
- All four `TestChaos_*` tests use `RunScenario` (no `runV2TestCase`, no inline `Pumba`)
- `setupChaos` no longer loads chains or filters by family; EVM lib loading stays at the test call site
- Uses `tcapi.AssertMessageOptions` and the V3 message pipeline

**RPC latency chaos (new):**
- `BuildNetemDelayCommand` / `LatencySpec` / `injectLatency` — Pumba `netem delay` command builder, mirroring the pattern already used in `load_test.go`
- `BlockchainContainer` / `BlockchainContainerForSelector` — resolve blockchain RPC container names by array index or chain selector (selector-based since `lib.Chains()` returns unspecified order)
- `ScenarioSpec.Latency` (`*LatencySpec`) — `RunScenario` dispatches latency vs outage via `injectChaos`; `Latency` takes precedence over `Outage` when both are non-nil
- `OutageSpec` and `LatencySpec` are now pointers on `ScenarioSpec` / `runEVMChaosScenario` so callers pass `nil` for the unused mode
- `TestChaos_RPCLatency` — injects 20000ms (20s) latency on both src+dest blockchain RPC containers for 1m, confirms V3 message finalizes offchain and executes on-chain despite the added latency

**RPC outage (blackout) — skipped:**
- Container resolvers (`BlockchainContainer`, `BlockchainContainerForSelector`) and `OutageSpec`/`InjectOutage` are in place, but the EVM test is not added because `ethclient.Client` (go-ethereum HTTP transport) has no reconnection logic. `CCIP17EVM` caches the client and all `abigen` bindings at env-setup time; after Pumba `stop --restart`, the `eventPoller` retries but the stale `http.Transport` connection pool keeps failing, so `ConfirmExecOnDest` hangs until timeout.
- TODO: add `TestChaos_RPCBlackout` once EVM devenv implements RPC reconnection (re-dial + re-bind on persistent failure). Solana devenv can use `OutageSpec` against its RPC node today since Solana clients reconnect natively. See `tcapi/chaos/README.md` for details.


build/devenv/tests/e2e/tcapi/chaos/scenario.go
```diff
@@ -0,0 +1,126 @@
...
+	ccv "github.com/smartcontractkit/chainlink-ccv/build/devenv"
+	"github.com/smartcontractkit/chainlink-ccv/build/devenv/cciptestinterfaces"
+	devenvcommon "github.com/smartcontractkit/chainlink-ccv/build/devenv/common"
+	"github.com/smartcontractkit/chainlink-ccv/build/devenv/tests/e2e/tcapi"
+	"github.com/smartcontractkit/chainlink-ccv/protocol"
+)
+
+// ScenarioSpec describes a chaos scenario: inject an outage, send a V3 message, assert
+// offchain progress, and optionally confirm execution on the destination chain.
+type ScenarioSpec struct {
```
makramkd, line 18: I haven't seen the full PR yet, but can we consolidate all these testing apis within `tcapi`? Main reason would be single point of truth for all testing APIs that is well documented, and we don't have to fish in different packages for e.g. Chaos or Load APIs.

makramkd (changes requested):

build/devenv/tests/e2e/chaos/evm_message.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/smartcontractkit/chainlink-ccv/protocol"
+	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
+)
+
+// EVMMessage holds V3 message fields and options hydrated for a simple EOA-receiver
+// send with the default committee CCV and executor on the source chain.
+type EVMMessage struct {
+	Fields  cciptestinterfaces.MessageFields
+	MsgOpts cciptestinterfaces.MessageOptions
+}
```
makramkd: Can we re-use the tcapi std test cases to avoid re-defining the same tests over again? See e.g. https://github.com/smartcontractkit/chainlink-ccv/blob/ec7a327d5e17efd21f20ae32ee9bd5d171921ac1/build/devenv/tests/e2e/tcapi/basic/v3.go#L268.

If for some reason that is unusable IMO we should generalize further tcapi if needed to accommodate these use cases.

build/devenv/tests/e2e/chaos/outage.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// ExecutorOutageDuration is used for executor and indexer outage scenarios.
+	ExecutorOutageDuration = 30 * time.Second
+)
+
+// OutageSpec describes a container stop outage injected via Pumba before a test
+// sends a message.
+type OutageSpec struct {
+	Duration time.Duration
+	Targets  []string
```
makramkd: Can we have a Go doc comment on this, its not super clear what should be in this slice just from reading this code.

build/devenv/tests/e2e/chaos/outage.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// LiteralSingle, when true with a single target, passes the container name to
+	// Pumba without ^$ anchors (aggregator nginx behavior).
+	LiteralSingle bool
```
makramkd: This seems odd, why is this needed? The comment doesn't really explain it that well.

build/devenv/tests/e2e/chaos/scenario.go
```diff
@@ -0,0 +1,126 @@
...
+// and optionally confirms execution on the destination. The caller must provide lib and
+// message fields; this package does not filter chains by family.
+func RunScenario(t *testing.T, ctx context.Context, spec ScenarioSpec) error {
+	cleanup, err := InjectOutage(ctx, spec.Outage)
+	if err != nil {
+		return fmt.Errorf("inject outage: %w", err)
+	}
+	t.Cleanup(cleanup)
+
+	chainMap, err := spec.Lib.ChainsMap(ctx)
```
makramkd: I feel like a lot of the boilerplate in this `RunScenario` is not needed if you simply invoke the appropriate std test case.

makramkd (changes requested):

build/devenv/tests/e2e/chaos/outage.go
huangzhen1997: This is seems to be a workaround for aggregator docker container name

huangzhen1997: Removed it.

build/devenv/tests/e2e/chaos_test.go
```diff
@@ -101,169 +62,134 @@ func TestChaos_VerifierFaultToleranceThresholdViolated(t *testing.T) {
...
-		AssertVerifierLogs:      false,
-		AssertExecutorLogs:      false,
+	runEVMChaosScenario(t, ctx, lib, src, dst, evmChaosConfig{
+		ConfirmExec: true,
+		Timeout:     5 * time.Minute,
+		ExecTimeout: 5 * time.Minute,
+	}, chaos.OutageSpec{
+		Duration: chaos.DefaultOutageDuration,
+		Targets:  executorTargets,
 	})
```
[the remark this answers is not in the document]
jadepark-dev: @huangzhen1997 This seems to be a valid point

build/devenv/tests/e2e/tcapi/v3message.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return err
+	}
+	testCtx, cleanupFn := NewTestingContext(ctx, aggregatorClient, indexerMonitor)
+	defer cleanupFn()
+
+	execTimeout := cfg.Run.ExecTimeout(cfg.ExecTimeout)
+	if execTimeout == 0 {
+		execTimeout = DefaultExecTimeout
+	}
```
jadepark-dev: Do you think we can avoid a silent default override? I wonder what the case is where we put 0 as a timeout?

build/devenv/tests/e2e/tcapi/basic/v3.go
```diff
@@ -65,82 +63,25 @@ func (tc *v3TestCase) Run(ctx context.Context) error {
...
-	v3Dst, err := tc.lib.V3Destination(ctx, tc.dst)
-	if err != nil {
-		return fmt.Errorf("destination chain %d does not support V3 message: %w", tc.dst, err)
-	}
-	l := zerolog.Ctx(ctx)
-	sendMessageResult, err := tcapi.SendV3Message(ctx, v3Src, v3Dst,
-		cciptestinterfaces.MessageFields{
-			Receiver: tc.receiver,
-			Data:     tc.msgData,
+	return tcapi.RunV3MessageLifecycle(ctx, tc.lib, tcapi.V3MsgConfig{
```
jadepark-dev: If this is only for reusability, can we defer this wrapping `RunV3MessageLifecycle`? Not a big fan of dumping everything in one config, which can nudge people to add random fields as they want

build/devenv/tests/e2e/chaos_test.go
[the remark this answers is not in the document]
jadepark-dev: nvm I just found original value in the chaos test file 
```outageDuration = 20 * time.Second```

File: build/devenv/tests/e2e/chaos/README.md [hunk 9]
```diff
@@ -0,0 +1,63 @@
+# e2e/chaos
+
+Shared helpers for CCIP devenv chaos tests: inject container outages via Pumba, resolve
```
jadepark-dev: @huangzhen1997 Looks like the chaos tests defined here cover:
- `TestChaos_AggregatorOutageRecovery`
- `TestChaos_VerifierFaultToleranceThresholdViolated`
- `TestChaos_AllExecutorsDown`
- `TestChaos_IndexerDown`

which makes sense to have platform-level failure cases

However, I wonder what the plan is around PRC failure scenarios(blackout, failure in multinode, abnormal latency)? Specifically with Solana, we should expect noise from RPC failure given that we have had many warrooms from it. I'm not sure if this is something we can provide from the CCV side, but curious to know what our approach is

cc. @ogtownsend

build/devenv/tests/e2e/tcapi/basic/v3.go
huangzhen1997: I was just about to revisit this, yes I agree, the initial goal was to dedup, but now it seems reduce the flexibility and readability, and more complex for the caller.

build/devenv/tests/e2e/tcapi/v3message.go
huangzhen1997: 0 is not allow because it means no timeout at all. For [reference](https://github.com/smartcontractkit/chainlink-ccv/blob/9002411014bbed57a7bb1a01669a559c01c74769/build/devenv/tests/e2e/chaos_test.go#L49), if the timeout is 0, stop --restart would stop the container and never restart it.

build/devenv/tests/e2e/chaos/README.md
huangzhen1997: This PR was just refactor existing chaos test coverage so we can add for Solana. But I agree with having rpc failure can be valuable. This chaos test technically supports inject outage to rpc node container, I've actually tested this locally, but test failed due to missing support of ethclient reconnection in evm devenv. It's out of scope here to support for EVM, but technically this is doable with existing chaos test outage injection when we add the test for Solana.

build/devenv/tests/e2e/tcapi/chaos/scenario.go
```diff
@@ -0,0 +1,141 @@
...
+			return nil, fmt.Errorf("inject latency: %w", err)
+		}
+		return cleanup, nil
+	}
+	cleanup, err := InjectOutage(ctx, spec.Outage)
+	if err != nil {
+		return nil, fmt.Errorf("inject outage: %w", err)
+	}
+	return cleanup, nil
+}
```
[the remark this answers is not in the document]
huangzhen1997: when both nil should return error as the function failed to injectChaos

build/devenv/tests/e2e/tcapi/basic/v3.go
```diff
@@ -205,6 +205,28 @@ func loadV3Env(ctx context.Context, lib ccv.Lib, src, dst uint64) (v3Env, bool)
 	return env, true
 }
 
+// ResolveEOAReceiverDefaultVerifier resolves the EOA receiver on dst and the default
+// committee CCV and default executor on src for a minimal V3 send.
+func ResolveEOAReceiverDefaultVerifier(ctx context.Context, lib ccv.Lib, src, dst uint64) (protocol.UnknownAddress, []protocol.CCV, protocol.UnknownAddress, error) {
```
ogtownsend: nit: seems to do more than just receiver and verifier, maybe call it something more high level like `ResolveV3SendRequiredAddresses()` or `GetMinSetV3SendAddresses()` 🤷

build/devenv/tests/e2e/tcapi/chaos/containers.go
```diff
@@ -0,0 +1,169 @@
...
+	"fmt"
+	"strings"
+
+	chain_selectors "github.com/smartcontractkit/chain-selectors"
+
+	ccv "github.com/smartcontractkit/chainlink-ccv/build/devenv"
+	devenvcommon "github.com/smartcontractkit/chainlink-ccv/build/devenv/common"
+	"github.com/smartcontractkit/chainlink-ccv/build/devenv/services/committeeverifier"
+)
+
```
ogtownsend, line 18: nit: add some brief toplevel documentation about what this file does/is responsible for, seems like it's main purpose is to provide easy access to test container names but doesn't actually manipulate container state or lifecycles

build/devenv/tests/e2e/tcapi/chaos/outage.go
```diff
@@ -0,0 +1,75 @@
...
+// sends a message.
+type OutageSpec struct {
+	Duration time.Duration
+	// Targets are normalized Docker container names (leading "/" stripped) that Pumba
+	// will stop during the outage.
+	Targets []string
+}
+
+// injectOutage starts a Pumba container stop for the given spec and returns a
+// cleanup function that tears down the Pumba sidecar.
```
ogtownsend: How does it work underneath? Does this mean a separate pumba container is spun up alongside a CCIP container (verifier, executor, etc) and sends it a signal to shut down, then is able to restart it later on (if desired) somehow?

build/devenv/tests/e2e/tcapi/chaos/README.md
```diff
@@ -0,0 +1,76 @@
...
+| `ExecutorContainersForDest(cfg, destSelector, qualifier)` | Dest-specific executor (cross-family devenvs) |
+| `IndexerContainer(cfg, index)` | Indexer outage |
+| `BlockchainContainer(cfg, index)` | Blockchain RPC outage by array index |
+| `BlockchainContainerForSelector(cfg, selector)` | Blockchain RPC outage by chain selector |
+
+Container names are normalized from env-out (`Out.ContainerName`, leading `/` stripped).
+
+## Limitations & TODOs
+
+### RPC outage (blackout) — TODO
```
makramkd: Can you create a ticket for this?

build/devenv/tests/e2e/tcapi/basic/v3.go
makramkd: I see this being used in this file only, at least in this PR, and its referenced in the README for the chaos pkg. Is it being used by the Solana chaos tests?

If it is, one other thought is whether this should be in `basic` or if we should create a `common` package of sorts inside tcapi. Not a blocker for this PR though.

makramkd (dismissed):
🚀

build/devenv/tests/e2e/tcapi/basic/v3.go
huangzhen1997: Yeah was going to rename but forgot, good catch!

build/devenv/tests/e2e/tcapi/chaos/outage.go
huangzhen1997: Pumba is a short-lived sidecar that issues docker stop + auto-restart via docker api, and the cleanup callback will remove thesidecar

build/devenv/tests/e2e/tcapi/chaos/README.md
huangzhen1997: https://smartcontract-it.atlassian.net/browse/CCIP-12571

build/devenv/tests/e2e/tcapi/basic/v3.go
huangzhen1997: I moved it along with some helpers to tcapi/resolve.go

tt-cll (approved):

ogtownsend (approved):

PR #1234, opened 2026-07-03
extract bootstrap secrets to env vars

nicolasgnr:
## Summary

This PR adds support for resolving sensitive configuration values from environment variables instead of requiring them to be stored in the TOML configuration file `bootstrap.toml`.

Specifically, the database URL and keystore password can now be provided via environment variables, making it easier to manage secrets securely. If both a value and its corresponding environment variable are configured, the value defined in the configuration file takes precedence.

## Changes

* Added `GetURL()` to `DBConfig` to resolve the database URL from either:

  * `URL` (highest precedence)
  * `URLEnvVar`
* Added `GetPassword()` to `KeystoreConfig` to resolve the keystore password from either:

  * `Password` (highest precedence)
  * `PasswordEnvVar`
* Refactored validation to use the new getter methods, centralizing configuration resolution and removing duplicated logic.
* Updated the bootstrap flow to use the new resolution methods.
* Added comprehensive tests covering:

  * Direct configuration values
  * Configuration precedence over environment variables
  * Environment variable fallback
  * Missing environment variables
  * Missing configuration from both sources

## Motivation

Secrets such as database credentials and keystore passwords should not be committed to or stored in plaintext configuration files. This change allows deployments to inject these values securely through environment variables while maintaining backward compatibility with existing configurations.

This allow as to move from

```toml
[jd]
server_wsrpc_url = "wsrpc-job-distributor.sh"
server_csa_public_key = "41dcc9f2d7....."

[keystore]
password = "sensitive info"

[db]
url = "sensitive info"
[server]
listen_port = 8080
```

to

```toml
[jd]
server_wsrpc_url = "wsrpc-job-distributor.sh"
server_csa_public_key = "41dcc9f2d7....."

[keystore]
password_env_var = "KEYSTORE_PASSOWORD_ENV_VAR"

[db]
url_env_var = "CL_DATABASE_URL"
        
[server]
listen_port = 8080
```

This change is fully backward compatible. Existing configurations that specify password and url directly continue to work unchanged.

In the future, once everything have migrated, we can remove the plaintext secret fields from the configuration, making environment variables the only supported mechanism for supplying sensitive values.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -352,12 +352,22 @@ func buildUpdateNodeRequest(
 
 // startWithJDLifecycle initializes all components required for the JD lifecycle manager and starts it.
 func (b *Bootstrapper) startWithJDLifecycle(ctx context.Context) error {
-	db, err := connectToDB(ctx, b.config.DB.URL)
+	dbURL, err := b.config.DB.GetURL()
+	if err != nil {
+		return err
+	}
+
+	db, err := connectToDB(ctx, dbURL)
 	if err != nil {
 		return fmt.Errorf("failed to connect to bootstrapper database: %w", err)
 	}
 
-	keyStore, csaSigner, err := initializeKeystore(ctx, b.lggr, db, b.config.Keystore.Password, b.keys)
+	keyStorePass, err := b.config.Keystore.GetPassword()
+	if err != nil {
+		return err
+	}
+
+	keyStore, csaSigner, err := initializeKeystore(ctx, b.lggr, db, keyStorePass, b.keys)
 	if err != nil {
 		return fmt.Errorf("failed to initialize keystore: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


bukata-sa (dismissed):

huangzhen1997 (dismissed):

makramkd (changes requested):
> Secrets such as database credentials and keystore passwords should not be committed to or stored in plaintext configuration files. This change allows deployments to inject these values securely through environment variables while maintaining backward compatibility with existing configurations.

I disagree w/ the motivation behind this PR. The chainlink node has been using a secrets.toml since v2 (4+ years in production at this point). Its pretty straightforward to set up a templated TOML file that uses secrets for its values for k8s deployments, I believe Chainlink node operators have been doing it for years now.

The `bootstrap.toml` config will never be sent by Chainlink, its entirely node operator provided. 

This change also violates the "stay consistent w/ Chainlink where reasonable" principle, creating more complexity in configuration isn't the right path forward, keeping things simple and consistent where applicable is better.

PR #1238, opened 2026-07-06
[CCIP-12096] bootstrap: split secrets out of monolithic config

makramkd:
## Description

The bootstrapper's operator config is a single TOML file that mixes credential-bearing sections (`[db]`, `[keystore]`) with non-sensitive ones (`[jd]`, `[server]`, `[[chains]]`, `[monitoring]`). Operators asked to deploy the two halves from different sources, the non-secret half from a ConfigMap, the secrets from a Secret with its own lifecycle and RBAC, which the monolithic file can't support without routing credentials through the non-secret delivery path.

This splits the config by sensitivity into a non-secret `config.toml` and a `secrets.toml`, **backwards-compatibly**: apps already deployed with the monolithic file keep working untouched.

Design:

- The partition lives in the type system as two embedded halves: `Config = struct { NonSecretConfig; Secrets }`. Because they're embedded (anonymous, not nested), TOML flattens their sections to the top level, so a monolithic file decodes into `Config` exactly as before, and each half can be marshalled independently to produce one of the two split files. This makes the partition a single source of truth.
- `LoadAndValidateConfig` now takes an ordered list of paths and decodes them into one struct: `config.toml` first, then `secrets.toml` overlaid on top, so the secrets file wins for any section it defines. Validation runs once on the merged struct, so a required section is checked regardless of which file supplied it.
- A new `BOOTSTRAPPER_SECRETS_PATH` env var (default `/etc/bootstrap/secrets.toml`) points at the secrets file. It's optional at the file level. Absence is skipped, which is what keeps a legacy monolith valid. Secrets loading is JD-mode only; static mode (token verifier) is untouched.
- devenv implements the split: the committee verifier and executor now emit and mount both a `config.toml` and a `secrets.toml`, so e2e exercises the merge path in real containers.

Migration: the monolithic layout is deprecated and slated for removal once operators cut over. No deprecation warning ships in this change.

## Testing

- `go test ./bootstrap/...` 
- `go test ./build/devenv/services/...` 

## Checklist

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -36,6 +36,12 @@ const (
 	ConfigPathEnv     = "BOOTSTRAPPER_CONFIG_PATH"
 	DefaultConfigPath = "/etc/config.toml"
 
+	// SecretsPathEnv names the env var pointing at the bootstrap secrets file, which carries the
+	// credential-bearing [db] and [keystore] sections. The default is namespaced
+	// under /etc/bootstrap/ to avoid the /etc/config.toml collision that DefaultConfigPath suffers.
+	SecretsPathEnv     = "BOOTSTRAPPER_SECRETS_PATH"
+	DefaultSecretsPath = "/etc/bootstrap/secrets.toml" //nolint:gosec // G101: this is a file path, not a credential
+
 	defaultStartupTimeout  = 10 * time.Second
 	defaultShutdownTimeout = 10 * time.Second
 )
@@ -125,6 +131,7 @@ type Bootstrapper struct {
 
 	// bootstrapper component configs
 	configPath       string
+	secretsPath      string
 	config           *Config
 	lifecycleManager *lifecycle.Manager
 	infoServer       *infoServer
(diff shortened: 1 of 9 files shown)


bootstrap/bootstrap.go
```diff
@@ -610,6 +652,18 @@ func WithBootstrapperConfigPath(path string) Option {
...
+// looks for the path in the BOOTSTRAPPER_SECRETS_PATH environment variable, and if that is not set,
+// it defaults to DefaultSecretsPath. The secrets file is optional: if the resolved path does not
+// exist, it is skipped (which is how a legacy monolithic config remains valid). Applies in JD mode
+// only.
+func WithBootstrapperSecretsPath(path string) Option {
+	return func(b *Bootstrapper) error {
+		b.secretsPath = path
+		return nil
+	}
+}
```
makramkd, line 665: I'd eventually want to remove WithBootstrapperSecretsPath and WithBootstrapperConfigPath, there's a ticket for it.

bootstrap/config.go
```diff
@@ -218,18 +248,24 @@ func (c *Config) validate(needsInfra bool) error {
...
-	if err != nil {
-		return fmt.Errorf("failed to read config file: %w", err)
-	}
-	// TODO switch to strict mode once config migration is over
-	err = parseTOML(string(tomlBytes), cfg, false)
-	if err != nil {
-		return fmt.Errorf("failed to parse config: %w", err)
+// static-TOML mode. Because validation runs on the merged struct, "is [db] present and valid" is
+// independent of which file supplied it.
+func LoadAndValidateConfig(paths []string, cfg *Config, needsInfra bool) error {
```
nicolasgnr, line 253: Do we expect to have other paths apart from config & secret tomls?

makramkd: Right now, no expectation.

changelog/2026-07-07_bootstrap_config_secret_split.md
```diff
@@ -0,0 +1,91 @@
+# Bootstrap config split into non-secret config + secrets files
+
+## Summary
+
+The bootstrapper's operator config can now be split by sensitivity into a
+non-secret `config.toml` and a `secrets.toml` (credentials only), so the two
+halves can be deployed from separate sources (e.g. a k8s ConfigMap vs a Secret).
+The legacy monolithic single-file layout still works unchanged. See
+`docs/adr/0008-bootstrap-secret-config-split.md`.
```
jadepark-dev: nit: missing ref here ;)

bootstrap/config.go
```diff
@@ -218,18 +242,24 @@ func (c *Config) validate(needsInfra bool) error {
...
+	for _, path := range paths {
+		tomlBytes, err := os.ReadFile(path) //nolint:gosec // G304: path is provided by trusted caller
+		if err != nil {
+			return fmt.Errorf("failed to read config file %q: %w", path, err)
+		}
+		// TODO switch to strict mode once config migration is over
```
jadepark-dev, line 259: Just an idea: probably we can at least force operators to clean up their monolith config file when `BOOTSTRAPPER_SECRETS_PATH` is intentionally set by checking duplicated keys. Currently secrets implicitly win, but I could picture an operator rotating credentials in the normal config and complaining that it doesn't work

jadepark-dev (dismissed):
Looks good to me, just a couple of minor comments

changelog/2026-07-07_bootstrap_config_secret_split.md
makramkd: Thanks! I'd like to start committing the ADR's at some point...

bootstrap/config.go
makramkd: I thought about that - I wanted to print a warning in that case instead of hard-failing, but unfortunately we don't have the beholder logger available currently when we load config (chicken and egg problem - we need to load the monitoring config in order to properly set up the logger). Maybe we can just use the `slog` package and log directly to stdout though.

bukata-sa (dismissed):

bootstrap/config.go
jadepark-dev: Yeah that could at least provide some additional context about the config issue

jadepark-dev (approved):

bukata-sa (approved):

carte7000 (approved):

PR #1240, opened 2026-07-07
verifier: load DB + aggregator creds from secrets file

makramkd:
## Description

Operators want the committee and token verifiers' sensitive configuration delivered as a mounted file (e.g. a k8s Secret) with its own lifecycle and RBAC, rather than as process environment variables.

This PR moves the credential-bearing values: the application storage DB URL (`CL_DATABASE_URL`) and, for the committee verifier, the per-aggregator HMAC credentials (`VERIFIER_AGGREGATOR_*`) — into an optional **verifier secrets file**, in a **backwards-compatible** way:

- The env vars still work unchanged. An absent file is never an error and falls back entirely to the environment (the existing prod path).
- When a value is set in both the file and the environment, the **file wins**, so operators can cut over safely: add the file, verify pickup, then strip the env vars.

Design decisions:

- **Verifier-owned secrets.toml file** `CL_DATABASE_URL` is the verifier's application DB, a *different* database than the bootstrap `[db]`. The token verifier also runs in static-TOML mode and never loads bootstrap secrets, yet uses `CL_DATABASE_URL`.
- **Schema is `[db].url` + `[[aggregators]]`** (array of tables with `secret_name` + `api_key` + `secret_key`), consistent with the Chainlink node secrets convention and the app config's own `[[aggregators]]`.
- **Per-app paths, shared loader:** `COMMITTEE_VERIFIER_SECRETS_PATH` / `TOKEN_VERIFIER_SECRETS_PATH`.
- **New leaf package `verifier/pkg/vsecrets`** holds the schema, constants, and loader so `commit`, `cmd/verifier`, and the devenv harness share one definition without pulling the CLI dependency tree into the devenv module.
- **Strict, fail-fast validation** of a *present* secrets.toml file.
- devenv is cut over to writing/mounting the file (dropping the env vars) so e2e exercises the real file path; the env-only path is covered by unit tests.

Follow-up (will create a separate JIRA ticket): delete the unused `CL_DATABASE_MAX_*` / `CONN_*` tuning knobs, which are set nowhere and always default.

## Testing

- `go test ./verifier/pkg/vsecrets/... ./verifier/pkg/commit/...
  ./cmd/verifier/...`
- `just lint fix`

## Checklist


cmd/verifier/committee/main.go
```diff
@@ -11,18 +11,27 @@ import (
...
 )
 
 func main() {
+	// Load the verifier secrets file once (DB URL + aggregator HMAC credentials); an absent file is
+	// fine and falls back to environment variables. Shared by both the CLI and the service.
+	secretsPath := vsecrets.ResolveSecretsPath(vsecrets.CommitteeVerifierSecretsPathEnv, vsecrets.DefaultCommitteeVerifierSecretsPath)
+	secrets, err := vsecrets.Load(secretsPath)
+	if err != nil {
+		panic(fmt.Sprintf("failed to load committee verifier secrets: %s", err.Error()))
+	}
```
makramkd: I'm wondering if this should be inside the factory - otherwise every chain-family binary would have to add it to the beginning of their main.go

cmd/verifier/common.go
```diff
@@ -20,13 +20,14 @@ import (
...
 )
 
 const (
-	// Database environment variables.
-	DatabaseURLEnvVar             = "CL_DATABASE_URL"
+	// Database environment variables. The URL is resolved via the verifier secrets package
+	// (vsecrets.DatabaseURLEnvVar); these tuning knobs remain plain env vars.
 	DatabaseMaxOpenConnsEnvVar    = "CL_DATABASE_MAX_OPEN_CONNS"
 	DatabaseMaxIdleConnsEnvVar    = "CL_DATABASE_MAX_IDLE_CONNS"
 	DatabaseConnMaxLifetimeEnvVar = "CL_DATABASE_CONN_MAX_LIFETIME"
```
makramkd: These aren't sensitive but I don't believe these are currently used anywhere - in the interest of simplifying things I think we should just have constants for these values and only pull them out into config when explicitly needed.

cmd/verifier/servicefactory.go
```diff
@@ -103,11 +109,13 @@ func (f *factory) Start(ctx context.Context, spec bootstrap.JobSpec, deps bootst
 	}
 	lggr.Infow("Resolved aggregators", "count", len(resolvedAggregators))
```
makramkd, line 117: Maybe we can read the secrets here instead?

verifier/pkg/commit/secrets.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"fmt"
+
+	"github.com/smartcontractkit/chainlink-ccv/protocol/common/hmac"
+)
+
+// buildHMACConfig validates an API key / secret pair and returns the client config. apiKeyLabel and
+// secretLabel name the source (env var names, or file field names) so error messages point the
+// operator at the right place regardless of where the credential came from.
+func buildHMACConfig(apiKey, secret, apiKeyLabel, secretLabel, aggLabel string) (*hmac.ClientConfig, error) {
```
makramkd: Probably doesn't have to be a new file - put it next to the call site in verifier/pkg/commit/config.go

verifier/pkg/commit/secrets_test.go
makramkd, line 1: Move these tests to config_test.go after moving buildHMACConfig to config.go

verifier/pkg/vsecrets/vsecrets.go
```diff
@@ -0,0 +1,163 @@
...
+const (
+	// DatabaseURLEnvVar is the legacy environment variable for the application storage DB URL, used
+	// as the fallback when the secrets file does not supply [db].url.
+	DatabaseURLEnvVar = "CL_DATABASE_URL"
+
+	CommitteeVerifierSecretsPathEnv     = "COMMITTEE_VERIFIER_SECRETS_PATH"
+	DefaultCommitteeVerifierSecretsPath = "/etc/committee-verifier/secrets.toml"
+
+	TokenVerifierSecretsPathEnv     = "TOKEN_VERIFIER_SECRETS_PATH"      //nolint:gosec // G101: env var name, not a credential
+	DefaultTokenVerifierSecretsPath = "/etc/token-verifier/secrets.toml" //nolint:gosec // G101: file path, not a credential
```
makramkd, line 33: Should the CommitteeVerifierSecretsPathEnv/DefaultCommitteeVerifierSecretsPath constants be defined in vsecrets or in the commit package? Similar question applies to the token counterpart constants.

huangzhen1997 (approved):

KodeyThomas (approved):

PR #1248, opened 2026-07-08
aggregator: load creds from secrets file

makramkd:
## Description

Moves the aggregator's credentials out of the process environment and into an optional, operator-owned secrets file (a mounted k8s Secret with its own lifecycle/RBAC), backwards-compatibly.

What moves into `secrets.toml`:
- `[storage].url` — the app storage DB URL (was `AGGREGATOR_STORAGE_CONNECTION_URL`)
- `[redis].password` — the rate limiter's redis password (was `AGGREGATOR_REDIS_PASSWORD`)
- `[[clients]]` — the per-client inbound HMAC `api_key`/`secret_key` pairs (were the `AGGREGATOR_API_KEY_*` / `AGGREGATOR_SECRET_*` env vars)

Key design choices (and where they differ from the verifier):
- **Backwards compatible**: an absent file is never an error and everything falls back to the existing env vars. When both supply a value, the file wins — enabling a staged cutover (add file, verify, strip env).
- **Replace-per-client precedence**: if the file supplies any pairs for a `client_id`, that set is authoritative and the env pairs for that client are ignored (no merge).
- **Scope**: every secret goes to the file for operator consistency, including the redis password. The non-secret redis knobs (`address`, `db`) stay on env vars for now (a follow-up should relocate them into the main config TOML, not the secrets file).

No breaking changes: the env-var path remains fully supported and is the only path exercised when no file is mounted.

## Testing

- Unit tests: `go test ./aggregator/pkg/secrets/... ./aggregator/pkg/model/...`
- Devenv E2E

## Checklist

### DIFF
--- a/aggregator/pkg/model/config.go
+++ b/aggregator/pkg/model/config.go
@@ -458,14 +458,14 @@ type ClientConfig struct {
 	Enabled     bool             `toml:"enabled"`
 	ClientID    string           `toml:"clientId"`
 	// resolvedPairs holds credentials supplied for this client by the aggregator secrets file. When
-	// non-nil it replaces the env-var (APIKeyPairs) source entirely for this client (replace-per-client
-	// file-wins). It is unexported so struct logging (e.g. the "Loaded configuration" log) can never
+	// non-nil it replaces the env-var (APIKeyPairs) source entirely for this client.
+	// It is unexported so struct logging (e.g. the "Loaded configuration" log) can never
 	// leak the plaintext credentials it carries. Populated by AggregatorConfig.ResolveSecrets.
 	resolvedPairs []resolvedAPIKeyPair
 }
 
 // effectivePairs returns the credentials in force for this client: the secrets-file-resolved pairs
-// when present (replace-per-client), otherwise the legacy env-var pairs. This is the single source
+// when present, otherwise the legacy env-var pairs. This is the single source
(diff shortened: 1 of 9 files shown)


carte7000 (dismissed):

aggregator/pkg/model/config.go
```diff
@@ -947,8 +1032,8 @@ func (c *AggregatorConfig) loadRateLimiterRedisConfigFromEnvironment() error {
 	}
 	c.RateLimiting.Storage.Redis.Address = redisAddress
 
-	redisPassword := os.Getenv("AGGREGATOR_REDIS_PASSWORD")
-	c.RateLimiting.Storage.Redis.Password = redisPassword
+	// Password comes from the secrets file when present, otherwise AGGREGATOR_REDIS_PASSWORD.
```
huangzhen1997, line 1035: Have you consider move the `AGGREGATOR_REDIS_ADDRESS` and `AGGREGATOR_REDIS_DB` into aggregator config instead of env var?

huangzhen1997 (dismissed):

huangzhen1997 (approved):

aggregator/pkg/model/config.go
makramkd: Ah yeah, considered those but those aren't sensitive info, just plain URL and db name.

carte7000 (approved):

PR #1249, opened 2026-07-08
CCIP-12199 enable beholder auth

bukata-sa:
## Summary

  - Wire CSA key signer into Beholder setup for authenticated telemetry.
  - Split monitoring init out of `NewBootstrapper` into `initMonitoring(signer)`, called separately in each start path — with signer in JD mode, without in TOML mode.
  - Add a temporary no-op logger in `NewBootstrapper` so the struct is usable before monitoring init.
  - Extract `CSASigner` interface from concrete `csaSigner` struct; `NewCSASigner` now returns the interface.
  - Add `beholderSigner` adapter that bridges `keys.CSASigner` to `beholder.Signer`.

  ## Why

  Beholder auth requires the CSA key, which is only available after keystore init — after `NewBootstrapper` returns. Monitoring init must be deferred to start time when the signer is in hand.

  ## Testing
`just rebuild-all`
`ccv obs up`
Check that verifier metrics (standalone mode) now have proper csa_public_key instead of `not-specified`

## Checklist


common/monitoring/setup.go
```diff
@@ -64,6 +84,23 @@ func SetupBeholder(config BeholderConfig, metricViews []sdkmetric.View) error {
 	return nil
 }
 
+var _ beholder.Signer = (*beholderSigner)(nil)
+
+type beholderSigner struct {
+	signer keys.CSASigner
+}
+
+func (b *beholderSigner) Sign(_ context.Context, keyID string, data []byte) ([]byte, error) {
```
makramkd: Does keyID correspond to the public key we give in the auth config above? AuthPublicKeyHex part

bootstrap/keys/csa.go
```diff
@@ -11,33 +11,23 @@ import (
...
 	publicKey crypto.PublicKey
 }
 
-var _ crypto.Signer = (*CSASigner)(nil)
+var _ CSASigner = (*csaSigner)(nil)
 
-// NewCSASigner returns a [crypto.Signer] for the named Ed25519 key in ks.
-func NewCSASigner(ctx context.Context, ks keystore.Keystore, keyName string) (*CSASigner, error) {
+// NewCSASigner returns a [CSASigner] for the named Ed25519 key in ks.
+func NewCSASigner(ctx context.Context, ks keystore.Keystore, keyName string) (CSASigner, error) {
```
makramkd: Is keyName the human readable name or the public key?

makramkd (dismissed):

makramkd (approved):

carte7000 (approved):

PR #1250, opened 2026-07-08
jd config toggle

tt-cll:
## Checklist


bootstrap/app.example.toml
makramkd, line 1: **issue**: I feel like its pretty odd to have the app-specific config example (a kind of documentation) in the `bootstrap/` directory. 

Related to these kinds of example configs, I'd like to have more comprehensive configuration documentation in this repo, namely, docs that are checked in CI to ensure no drift.

So IMO this kind of example should be omitted, as the BOOTSTRAPPER_MODE=local should be usable by either verifier or executor (and current example seems to be verifier).

bootstrap/bootstrap.go
```diff
@@ -42,10 +41,38 @@ const (
...
+	// mode: "jd" or "local". When unset it falls back to the per-service default (JD unless the
+	// service passed WithLocalModeDefault), so the committee verifier and executor default to JD but
+	// switch to local for free, while the token verifier defaults to local.
+	ModeEnv = "BOOTSTRAPPER_MODE"
+
+	// LocalConfigPathEnv names the env var pointing at the app config file read in local mode. The
+	// file is a plain app-config TOML (the same content JD would ship in a job's appConfig), which
+	// the bootstrapper wraps and hands to the service factory.
+	LocalConfigPathEnv     = "BOOTSTRAPPER_LOCAL_CONFIG_PATH"
+	DefaultLocalConfigPath = "/etc/bootstrap/app.toml"
```
makramkd: **issue**: in the spirit of moving away from env vars for reasons of documentation and consistency, I believe a more appropriate set up would be to have these in the base bootstrap config (not secrets - since no sensitive info):

```toml
# nit: not sure what a good name would be here 🤔 
[mode]
# analogue: "jd_app_config"
app_config_mode = "local_app_config" 
# nit: no default for this, should always be present if bootstrap_app_config_mode == "local_config"
# can be empty otherwise.
local_app_config_path = "/etc/myapp/config.toml" 
```

Some additional reasoning around the "app" prefix: bootstrap already has a "local" config file - I believe we need more disambiguation, o/w it would lead to confusion.

bootstrap/bootstrap.go
```diff
@@ -42,10 +41,38 @@ const (
...
+const (
+	// modeJD loads the app config from JD and runs the full job lifecycle. Default for services that
+	// support JD (committee verifier, executor).
+	modeJD mode = iota
+	// modeLocal runs without JD: it reads the app config from a local file and, when a [db]+[keystore]
+	// bootstrap config is provided, initializes a Postgres-backed keystore so the service can sign.
+	// The token verifier (which needs no keystore) and the committee verifier/executor (which do)
+	// both run under this one mode; the keystore is initialized only when configured.
+	modeLocal
+)
```
makramkd: **nit**: I would prefer constants here that map to exact values that need to be provided by the user running the app, so that things are more self documenting. Also the `AppConfig` prefix for the name to clearly document what AppConfigMode means

```golang
// AppConfigMode is the manner in which the bootstrapper loads application config.
// It is set by users via [mode].app_config_mode in the bootstrap config.toml.
type AppConfigMode string

const (
  // AppConfigModeJD indicates to the bootstrapper to initiate the JD lifecycle to fetch app configs.
  AppConfigModeJD AppConfigMode = "jd_app_config"
  // AppConfigModeLocal indicates to the bootstrapper to read a file for application config.
  // nit: maybe should be AppConfigModeFile instead?
  AppConfigModeLocal AppConfigMode = "local_app_config"
)
```

bootstrap/bootstrap.go
```diff
@@ -654,14 +758,22 @@ func WithKey(name, purpose string, keyType keystore.KeyType) Option {
...
+// WithLocalConfigPath sets the app-config file path read in local mode. If not set, the bootstrapper
+// looks for BOOTSTRAPPER_LOCAL_CONFIG_PATH and falls back to DefaultLocalConfigPath. Applies in
+// local mode only.
+func WithLocalConfigPath(path string) Option {
 	return func(b *Bootstrapper) error {
-		b.jdMode = true
+		b.localConfigPath = path
 		return nil
 	}
 }
```
makramkd: **issue**: I think its getting really confusing already w/ how these options play w/ config and env vars. In my mind I'd like to remove most of these functional options and make things entirely config-driven instead. The main reason being that passing new / changing functional options required a complete rebuild/republish of a docker image, whereas a config driven approach just requires a config change.

So:
* Ditch as many env vars as possible, unless absolutely necessary (BOOTSTRAPPER_CONFIG_PATH and BOOTSTRAPPER_SECRET_PATH is more than enough)
* Ditch as many of these functional options as possible, move to config.toml

In my mind it makes sense to clean up functional options as a pre-req if necessary to make this PR a bit cleaner - but will leave it to you to see if its necessary or a fast follow.

makramkd (dismissed):
LG overall but I would like the mode and other info to be provided via config.toml rather than env vars.

bootstrap/bootstrap.go
makramkd: Alternatively, maybe you can argue for top-level keys (i.e. no `[mode]`):

```toml
app_config_mode = "local_app_config"
local_app_config_path = "/etc/myapp/config.toml"
```

bootstrap/job_test.go
```diff
@@ -51,7 +51,7 @@ type testAppConfig struct {
 	Value int    `toml:"value"`
 }
 
-// Exercises deprecated GetGenericConfig before it's fully removed.
+//nolint:staticcheck // SA1019: exercises deprecated GetGenericConfig/GenericConfig before it's fully removed
```
tt-cll: This was pre-existing on main. Don't know how it passed CI

bootstrap/bootstrap.go
```diff
@@ -466,24 +423,101 @@ func (b *Bootstrapper) startWithJDLifecycle(ctx context.Context) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to initialize keystore: %w", err)
+		}
+	case dbURL != "" || ksPassword != "":
+		// Exactly one of [db]/[keystore] is set — almost certainly a misconfiguration. Warn loudly:
+		// running keystore-less here would surface later as a confusing nil-keystore failure in a
+		// signing service.
+		b.lggr.Warnw("local mode: bootstrap config sets only one of [db].url / [keystore].password; "+
+			"both are required to initialize the keystore, so the service will run keystore-less",
+			"hasDBURL", dbURL != "", "hasKeystorePassword", ksPassword != "")
```
makramkd, line 457: **question**: logger init is kinda finicky in bootstrap - are we sure that `b.lggr` is non-nil here? I recall `initMonitoring` also does logger init.

bootstrap/bootstrap.go
```diff
@@ -466,24 +423,101 @@ func (b *Bootstrapper) startWithJDLifecycle(ctx context.Context) error {
...
+	}
+
+	if err := b.initMonitoring(csaSigner); err != nil {
+		return fmt.Errorf("failed to initialize monitoring: %w", err)
+	}
```
makramkd, line 462: **nit**: an ordering nit (I could be wrong), can `initMonitoring` run prior to the DB/Keystore connectivity above? Or, I guess, it needs the CSA key from the keystore first before it can hook it up to beholder eh?

bootstrap/job_test.go
makramkd: There's something finicky w/ the linter on staticcheck...

cmd/verifier/token/main.go
```diff
@@ -39,15 +39,14 @@ func main() {
 		cmd.RunCCVCLI(os.Args[1:], vsecrets.TokenVerifierSecretsPathEnv, vsecrets.DefaultTokenVerifierSecretsPath)
 		return
 	}
-	configPath := os.Getenv("TOKEN_VERIFIER_CONFIG_PATH")
-	if configPath == "" {
-		configPath = "/etc/config.toml"
-	}
```
makramkd, line 45: **question**: would this be a breaking change? Good to document if so, so that the next token verifier deployment is updated w/ the appropriate.

build/devenv/services/committeeverifier/base.go
```diff
@@ -37,6 +37,10 @@ const (
 	DefaultVerifierMode    = services.Standalone
 
 	DefaultVerifierDBImage = "postgres:16-alpine"
+
+	// localAppConfigContainerPath is where the local-mode app-config file is mounted; it is written
+	// into the bootstrap config as local_app_config_path so the bootstrapper reads it there.
+	localAppConfigContainerPath = "/etc/bootstrap/app.toml"
```
makramkd:
```suggestion
	// localAppConfigContainerPath is where the local-mode app-config file is mounted; it is written
	// into the bootstrap config as local_app_config_path so the bootstrapper reads it there.
	localAppConfigContainerPath = "/etc/committe-verifier/app.toml"
```

build/devenv/services/committeeverifier/base.go
```diff
@@ -113,6 +117,12 @@ type Input struct {
 	// GeneratedJobSpecs[NodeIndex % numAggregators].
 	// Used in standalone mode. Set by generateVerifierJobSpecs in environment.go.
 	GeneratedConfig string `toml:"-"`
+
+	// LocalAppConfig is the plain app-config TOML mounted into the container in local mode
+	// (services.Local) — the committee verifier's commit.Config with blockchain_infos included, the
+	// same content JD would ship in a job's appConfig, no envelope. Callers build it (base.go cannot
```
makramkd, line 123: **note**: We wanna eventually axe blockchain_infos from the app config, its chain-specific config (e.g. /etc/evm/config.toml), wondering if we have a ticket for it...

build/devenv/services/common.go
```diff
@@ -21,6 +21,10 @@ type Mode string
 const (
 	Standalone Mode = "standalone"
 	CL         Mode = "cl"
+	// Local runs the service standalone without a Job Distributor: the bootstrapper reads the app
+	// config from a mounted job-spec file (BOOTSTRAPPER_MODE=local). Used by the CCV starter kit and
+	// local testing where JD is unavailable.
```
makramkd:
```suggestion
	// Local runs the service standalone without a Job Distributor: the bootstrapper reads the app
	// config from a mounted job-spec file (`app_config_mode = local` in bootstrap config.toml). 
	// Used by the CCV starter kit and local testing where JD is unavailable.
```

build/devenv/services/tokenVerifier.go
```diff
@@ -165,10 +165,14 @@ func NewTokenVerifier(in *TokenVerifierInput, blockchainOutputs []*blockchain.Ou
...
-	// Generate and write the bootstrap (operator) config, carrying monitoring from the generated config.
-	// The token verifier runs in static-TOML mode with no infra, so only the non-secret monitoring
-	// section is written (no secrets file); validation correctly skips infra checks.
-	bootstrapConfig, err := toml.Marshal(bootstrap.NonSecretConfig{Monitoring: in.Bootstrap.Monitoring})
+	// Generate and write the bootstrap (operator) config. The token verifier runs in local app-config
+	// mode with no infra ([db]/[keystore]), so the non-secret config carries only the mode selection,
+	// the app-config path, and monitoring (no secrets file); validation correctly skips infra checks.
+	bootstrapConfig, err := toml.Marshal(bootstrap.NonSecretConfig{
+		AppConfigMode:      bootstrap.AppConfigModeLocal,
+		LocalAppConfigPath: "/etc/token-verifier-app-config.toml",
```
makramkd: **super nit**: 
```suggestion
		LocalAppConfigPath: "/etc/token-verifier/config.toml",
```

makramkd (commented):
Mostly nits, looks good!

bootstrap/bootstrap.go
tt-cll: yeah it looks like `NewBootstrapper` always sets b.lggr

bootstrap/bootstrap.go
tt-cll: yep exactly. I think we're safe

cmd/verifier/token/main.go
tt-cll: good catch. going to add a changelog

makramkd (approved):

nvsriram (approved):

PR #1257, opened 2026-07-09
feat: support JD-free devenvs

tt-cll:
## Checklist

- [x] PR title follows Conventional Commits (see `CONTRIBUTING.md`); breaking changes marked with `!` / `BREAKING CHANGE:`


bootstrap/bootstrap.go
```diff
@@ -43,6 +44,13 @@ const (
 
 	defaultStartupTimeout  = 10 * time.Second
 	defaultShutdownTimeout = 10 * time.Second
+
+	// localConfigPollInterval and localConfigWaitTimeout bound the local-mode wait for a
+	// not-yet-present app config file (see startLocal). The timeout is generous because devenv
+	// bring-up (deploy contracts, generate config) can take several minutes before the file is
+	// written; on timeout the watcher gives up and the service simply never starts.
+	localConfigPollInterval = 2 * time.Second
+	localConfigWaitTimeout  = 15 * time.Minute
```
makramkd: **question**: this is because of how today, devenv brings up the containers "early" and only then ships the config via JD?

I guess the alternative here is that devenv startup logic would need to be tweaked - startup the containers early in the case of JD mode, later (when the app configs are generated) in the case of local mode?

I think its clear the complexity has to be _somewhere_ though. I'd vote for it being in devenv vs. the bootstrap, because I don't think that it makes a lot of sense to poll for a file eventually being available in the container - in real deployments this would never be the case, so its odd to support it first-class like this IMO.

tt-cll: moved to devenv

makramkd (approved):

carte7000 (approved):

PR #1270, opened 2026-07-14
feat(jd): stage replacement jobs before cutover (CCIP-11875)

tt-cll:
## Description

https://smartcontract-it.atlassian.net/browse/CCIP-11875

Stages and validates replacement JD jobs before stopping the active job, with rollback and idempotent resource cleanup. This keeps services available when replacement preparation fails and prevents resource leaks during cutover.

## Testing

Unit tests cover staged replacement ordering, rollback, readiness preservation, candidate cleanup, and idempotent registry and factory shutdown. Devenv covers regression

## Checklist

- [x] PR title follows Conventional Commits (see `CONTRIBUTING.md`); breaking changes marked with `!` / `BREAKING CHANGE:`


common/jd/lifecycle/manager.go
```diff
@@ -302,42 +318,45 @@ func (m *Manager) handleProposal(proposal *pb.ProposeJobRequest) (retErr error)
...
-			// The job will need to be re-proposed after fixing the error (whatever it may be).
-			m.lggr.Warnw("Failed to persist pending proposal", "error", err)
-			m.metrics.IncStepError(metricsCtx, stepSavePending, wasRunning)
-		} else {
-			m.lggr.Infow("Proposal persisted as pending", "proposalID", proposal.Id)
+	if wasRunning {
+		// Reject invalid application config while the old job is still serving. Validation is
+		// deliberately side-effect free, so it runs before the proposal is persisted and needs
+		// no rollback. Runtime resources are built by StartJob only after cutover.
+		if err := m.validateReplacement(proposal.Spec); err != nil {
```
makramkd, line 325: Nice ‼️

common/jd/lifecycle/manager.go
```diff
@@ -406,13 +425,31 @@ func (m *Manager) startJob(spec string) error {
...
+	return runner.ValidateJob(ctx, spec)
+}
+
+// rollbackReplacement is called after replacement cutover has been attempted. StopJob may return
+// an error after partially stopping the service, so both stop and start failures use the same
+// recovery: remove the pending record and restart the old spec.
+func (m *Manager) rollbackReplacement(newProposalID string, cause error, oldJob *store.Job) error {
+	ctx, cancel := context.WithTimeout(context.Background(), m.controlPlaneTimeout)
+	defer cancel()
 	metricsCtx := context.Background()
```
makramkd, line 447: **issue**: not part of your PR (seems like I missed it in an older review) `context.Background()` is always a bit fishy especially when there's I/O involved, I guess we should either use plain `ctx` or create a new one just for the metrics (but that seems overkill).

makramkd (approved):
Awesome! I think this is a genuine resilience improvement, especially on incorrect job specs, which is a real failure mode :-)

common/jd/lifecycle/manager.go
AmrMohamedRezk: If this fails validation should we call `RejectJob`  ?

nvsriram (approved):

PR #1281, opened 2026-07-21
feat(devenv)!: use narrower chain interfaces in tcapi

makramkd:
## Description

`tcapi.SendV3Message` took `src, dst cciptestinterfaces.CCIP17` and immediately type-asserted them down to the handful of methods it actually used. That forced any chain family wanting to run `tcapi`'s V3 message tests to implement the *entire* `CCIP17` interface (`Chain` + `Observable`, 13 methods), including deployment/funding-bootstrap methods (`NativeBalance`, `TransferNative`, `ManuallyExecuteMessage`, `ExposeMetrics`, etc.) that have nothing to do with sending or receiving a V3 message. 

The same over-coupling was scattered across the other `tcapi` V3 test cases (`ConfirmExecOnDest`, `GetTokenBalance`, `GetSenderAddress`, `GetMaxDataBytes`, `GetEOAReceiverAddress` all called directly on `CCIP17`-typed values pulled from `ChainsMap()`), and there was no registration path for a chain family to plug in with anything less than a full `CCIP17` implementation.

## Solution

- Added narrow compound interfaces `cciptestinterfaces.V3Source` (`MessageV3Source` + `ChainAsSource`) and `V3Destination` (`MessageV3Destination` + `ChainAsDestination`). These are exactly the capabilities every V3 test case needs, nothing more.
- `SendV3Message` now takes `src V3Source, dst MessageV3Destination` directly; no internal type assertions remain.
- `chainreg.Registration` gains two new **optional** fields, `V3SourceFactory`/`V3DestinationFactory`, so a family can register V3 send/receive support without implementing the full `ImplFactory`/`CCIP17` surface. EVM registers both (`evm.NewV3Source`/`NewV3Destination`) as the reference implementation for Solana/Canton/Aptos/Stellar to copy.
- `ccv.Lib` gains `V3Source`/`V3Destination` lookup methods that check the family's registered factory first, falling back to asserting the existing `ChainsMap()`-sourced `CCIP17` value. Existing families (EVM, and the extra-args-only Canton/Solana stubs) keep working unchanged.
- Capabilities only *some* test cases need (`GetTokenBalance`, `GetSenderAddress`, `GetMaxDataBytes`) were deliberately **not** bundled into `V3Source`/`V3Destination`. 
    - They're factored into small standalone interfaces (`TokenBalanceReader`, `SenderAddressProvider`, `MaxDataSizeProvider`, all now embedded in the full `Chain` interface for backward compatibility) and type-asserted at the specific call site that needs them — mirroring the existing `ProgressableChain`/`ReorgableChain`/`TokenConfigProvider` pattern in this package, so a family isn't forced to implement token-balance reporting just to run the basic send/confirm tests.
- Removed a dead `TestingContext.Impl` field (never read anywhere in `tcapi`) and its `NewTestingContext` parameter, eliminating the last unnecessary `ChainsMap()` fetch in every `tcapi` call site.

Net effect: `build/devenv/tests/e2e/tcapi` (and its `testutils/tokenpool` consumer) no longer references `cciptestinterfaces.CCIP17` anywhere.

## Checklist


build/devenv/tests/e2e/tcapi/types.go
```diff
@@ -86,42 +86,30 @@ type SendArgs struct {
 // SendV3Message builds and sends a V3 message using BuildV3ExtraArgs, BuildChainMessage, and SendChainMessage.
 func SendV3Message(
 	ctx context.Context,
-	src, dst cciptestinterfaces.CCIP17,
+	src cciptestinterfaces.V3Source,
+	dst cciptestinterfaces.MessageV3Destination,
 	destSelector uint64,
```
jadepark-dev: Loving direct params, but I found myself wondering why we get a full interface for source vs. only message interface for dest. wdyt passing `V3Destination` for dst? Probably less lean since we don't need dest methods here, but `destSelector` will be available for free

build/devenv/tests/e2e/testutils/tokenpool/token_transfer.go
```diff
@@ -47,16 +47,11 @@ func transferTokens(
 ) {
 	t.Helper()
 
-	chainMap, err := lib.ChainsMap(t.Context())
-	require.NoError(t, err, "get chains map")
-
```
jadepark-dev, line 52: 🚀

build/devenv/tests/e2e/tcapi/types.go
makramkd: What you suggest is also an improvement because it makes the method easier to use - harder to pass `dst` where `dst.ChainSelector() != destSelector`.

makramkd: Implemented. Thanks 🙏

jadepark-dev (approved):

bukata-sa (approved):

PR #1284, opened 2026-07-21
feat: Use redis for heartbeat config

carte7000:
Use redis for heartbeat config

### DIFF
--- a/aggregator/pkg/heartbeat/storage.go
+++ b/aggregator/pkg/heartbeat/storage.go
@@ -39,15 +39,15 @@ type RedisStorage struct {
 	ttl       time.Duration
 }
 
-func NewStorageFromConfig(l logger.SugaredLogger, c *model.AggregatorConfig) Storage {
-	if c.Heartbeat.StoreType == model.HeartbeatStoreTypeRedis {
+func NewStorageFromConfig(l logger.SugaredLogger, c model.HeartbeatConfig) Storage {
+	if c.StoreType == model.HeartbeatStoreTypeRedis {
 		client := redis.NewClient(&redis.Options{
-			Addr:     c.Heartbeat.Redis.Address,
-			Password: c.Heartbeat.Redis.Password,
-			DB:       c.Heartbeat.Redis.DB,
+			Addr:     c.Redis.Address,
+			Password: c.Redis.Password,
+			DB:       c.Redis.DB,
 		})
-		l.Infof("Using Redis heartbeat storage at %s with key prefix '%s' and TTL %s", c.Heartbeat.Redis.Address, c.Heartbeat.Redis.KeyPrefix, c.Heartbeat.Redis.TTL)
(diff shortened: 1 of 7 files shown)


aggregator/pkg/model/config.go
```diff
@@ -426,6 +461,39 @@ type AggregatorConfig struct {
 	MaxCommitVerifierNodeResultRequestsPerBatch int `toml:"maxCommitVerifierNodeResultRequestsPerBatch"`
 }
 
+type HeartbeatConfig struct {
+	// StoreType selects the storage backend for heartbeat data; currently "memory" or "redis".
+	StoreType HeartbeatStoreType `toml:"storeType"`
+	// Redis configuration (only used when StoreType is "redis")
+	Redis *HeartbeatRedisConfig `toml:"redis,omitempty"`
```
makramkd, line 468: This currently isn't showing up in config.documented.toml; For the sake of documenting this though,  can you update aggregatorConfigInstance in tools/configdoc/registry/registry.go to add something like:

```golang
// in aggregatorConfigInstance
var heartbeatRedisConfig aggregator.HeartbeatRedisConfig
heartbeatRedisConfig.SetDefaults()

// add to the c value
Heartbeat: aggregator.HeartbeatConfig{
  StoreType: "",
  Redis:     &heartbeatRedisConfig,
},
```

If you run `just config-docs` it'll tell you:

```
configdoc: aggregator: HeartbeatRedisConfig.KeyPrefix: missing doc comment
HeartbeatRedisConfig.TTL: missing doc comment
```

Once you add the doc comments to those fields, you can run it again and it should add this to the documented config toml:

```toml
# heartbeat configures the heartbeat monitoring.
[heartbeat]
  # storeType selects the storage backend for heartbeat data; currently "memory" or "redis".
  storeType = ""
  # redis configuration (only used when StoreType is "redis") <----- NEW
  [heartbeat.redis]
    # keyPrefix is the key prefix to use for the heartbeat values.
    keyPrefix = "heartbeat"
    # ttl is the time to live of the heartbeat values.
    ttl = "168h0m0s"
```

makramkd: That said, this shouldn't have to be detected by humans, I'll create a bug ticket for the config doc generator.

aggregator/pkg/heartbeat/storage.go
```diff
@@ -36,6 +39,20 @@ type RedisStorage struct {
 	ttl       time.Duration
 }
 
+func NewStorageFromConfig(l logger.SugaredLogger, c *model.AggregatorConfig) Storage {
```
makramkd: nit: looks like c.Heartbeat is only ever accessed, would it make sense to pass that in directly instead (makes the function signature more honest)

makramkd (approved):

kirtan-amin (approved):

PR #1285, opened 2026-07-21
feat: Backpressure improvements

carte7000:
This PR is changing 2 behaviour in the aggregation.

1. It eliminate block time when the client channel is full. Previously the channel manager would block for X seconds before throwing the `ErrAggregationChannelFull` error. With this change it will throw it instantly. This prevent aggregator keeping request open when it is already under high load.

2. Another change we made is that if another client sent a request for the same message and that this message have not been checked for aggregation we will not duplicate the aggregationRequest. This save some resources since we almost always have 16 NOPs sending their verification around the same time. If aggregator is under high load and processing is tight this effect will be amplified.

We also add a validation in the config now since the max batch size for verification need to be at least the size of the channel otherwise large batch would always be rejected. The default channel size will default to 2x the max batch size

### DIFF
--- a/aggregator/pkg/aggregation/aggregation_channel_manager.go
+++ b/aggregator/pkg/aggregation/aggregation_channel_manager.go
@@ -29,7 +29,7 @@ func NewChannelManager(keys []model.ChannelKey, bufferSize int) *ChannelManager 
 	manager := &ChannelManager{
 		clientChannel:       make(map[model.ChannelKey]chan aggregationRequest),
 		clientOrder:         make([]model.ChannelKey, 0, len(keys)),
-		AggregationChannel:  make(chan aggregationRequest),
+		AggregationChannel:  make(chan aggregationRequest, len(keys)),
 		wakeUp:              make(chan struct{}, 1),
 		currentlyQueued:     make(map[string]struct{}),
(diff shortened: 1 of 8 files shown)


aggregator/pkg/aggregation/aggregation_channel_manager.go
```diff
@@ -16,14 +17,21 @@ type ChannelManager struct {
...
 }
 
 func NewChannelManager(keys []model.ChannelKey, bufferSize int) *ChannelManager {
 	manager := &ChannelManager{
 		clientChannel:      make(map[model.ChannelKey]chan aggregationRequest),
 		clientOrder:        make([]model.ChannelKey, 0, len(keys)),
-		AggregationChannel: make(chan aggregationRequest, len(keys)),
+		AggregationChannel: make(chan aggregationRequest),
 		wakeUp:             make(chan struct{}, 1),
+		currentlyQueued:    make(map[string]struct{}),
```
bukata-sa: lock is not initialized

carte7000: Fixed

makramkd (approved):

kirtan-amin (approved):

PR #1289, opened 2026-07-22
feat(devenv)!: expose src/dest tx hashes in tcapi run results

jadepark-dev:
## Description

Exposes the on-chain source/destination transaction identifiers from tcapi test runs, and fixes token-transfer receiver handling so cross-family (e.g. EVM to Solana) token-only transfers work using tcapi.

**feat: transaction ids in run results**
- Add `cciptestinterfaces.SentEnvelope` / `ExecEnvelope`: the chain-agnostic event paired with an opaque `protocol.ByteSlice` `TxID` (EVM tx hash, Solana signature, etc.). Each chain names the value natively in its own impl (EVM `txHash`) and maps it into `TxID` at the interface boundary.
- Add `tcapi.RunResult{Src, Dest}` and make `TestCase.Run` return it (`Run(ctx) (RunResult, error)`). This lets integration-owned tests inspect the source send tx and destination exec tx (e.g. to read pool events from the exact tx).
- Populate the previously-unset `ExecutionStateChangedEvent.SourceChainSelector` in the EVM off-ramp poller.
- Breaking: `ConfirmExecOnDest` now returns `(ExecEnvelope, error)` and `TestCase.Run` now returns `(RunResult, error)`; `SendV3Message` gains a tx-id return. All in-repo call sites updated (`_` where the id isn't needed).

**fix: token receiver**
- Token-transfer balance assertions now target the resolved token-receiver address (`tc.tokenReceiver`), falling back to the message receiver when no token receiver is supplied (during hydrate).
- For a token-only transfer that carries a separate token receiver, the message receiver is left empty so destination chains that treat a non-zero receiver as an arbitrary-message delivery (e.g. Solana, where `is_arbitrary_message` is `receiver != 0`) don't attempt a receiver CPI for a transfer that has no payload.

## Testing
- https://github.com/smartcontractkit/chainlink-ccip-solana/pull/272

## Checklist

- [x] Crosslink related PRs (in this or other repositories)


build/devenv/cciptestinterfaces/interface.go
```diff
@@ -89,6 +89,14 @@ type MessageSentEvent struct {
 	VerifierBlobs  [][]byte
 }
 
+// SentEnvelope preserves the chain-agnostic sent event together with the
+// source transaction identifier. TxHash is opaque: for example, it is an EVM
+// transaction hash or a Solana signature.
+type SentEnvelope struct {
+	TxHash protocol.ByteSlice
```
makramkd: Question: previous iteration called this TxID, would that be more agnostic?

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -147,7 +163,7 @@ type Chain interface {
 	// ConfirmSendOnSource waits until exactly one CCIPMessageSent event is emitted on-chain for the specified destination chain, identified by sequence number or message ID.
 	ConfirmSendOnSource(ctx context.Context, to uint64, key MessageEventKey, timeout time.Duration) (MessageSentEvent, error)
 	// ConfirmExecOnDest waits until exactly one ExecutionStateChanged event is emitted on-chain for the specified source chain, identified by sequence number or message ID.
-	ConfirmExecOnDest(ctx context.Context, from uint64, key MessageEventKey, timeout time.Duration) (ExecutionStateChangedEvent, error)
+	ConfirmExecOnDest(ctx context.Context, from uint64, key MessageEventKey, timeout time.Duration) (ExecutionStateChangedEvent, protocol.ByteSlice, error)
```
makramkd: Nit: named returns here would be good to make clear what the byteslice refers to

build/devenv/cciptestinterfaces/interface.go
```diff
@@ -351,7 +367,7 @@ type ChainAsDestination interface {
 	// ConfirmExecOnDest confirms that a CCIP message was executed on this chain.
 	// Implementation should support confirmation by either message ID or sequence number, passed in as the `MessageEventKey`.
 	// The timeout is the maximum duration to wait for the event to be emitted.
-	ConfirmExecOnDest(ctx context.Context, from uint64, key MessageEventKey, timeout time.Duration) (ExecutionStateChangedEvent, error)
+	ConfirmExecOnDest(ctx context.Context, from uint64, key MessageEventKey, timeout time.Duration) (ExecutionStateChangedEvent, protocol.ByteSlice, error)
```
makramkd: Ditto here re: returning envelope

build/devenv/tests/e2e/tcapi/types.go
```diff
@@ -46,6 +47,16 @@ type TestCase interface {
 	HavePrerequisites(ctx context.Context) bool
 }
 
+// ObservableTestCase is optionally implemented by test cases that expose the
+// source and destination transaction evidence produced by Run.
+type ObservableTestCase interface {
```
makramkd: Question: do we need a new interface? Why not a breaking change to the Run method directly?

build/devenv/cciptestinterfaces/interface.go
makramkd: Actually can we change this to return the envelope, instead of two separate return values?

jadepark-dev: Hmm yeah I was trying not to return custom wrapper type, but well I already did that for `RunResult`

build/devenv/cciptestinterfaces/interface.go
jadepark-dev: Sounds good, at some point I saw one TxID being used somewhere and pivoted

makramkd (approved):

huangzhen1997 (approved):

PR #1292, opened 2026-07-22
feat(evm): use production chain services in standalone accessors

tt-cll:
## Description

Replace the standalone EVM accessor's early-stage head tracker and direct transaction sender with chainlink-evm's production services.

- Use the production multi-node client, head tracker, head broadcaster, gas estimator, and TXM v2.
- Translate CCV's focused EVM configuration into chainlink-evm configuration while preserving RPC failover, HTTP polling, WebSocket subscriptions, and CCV's finality depth.
- Defer RPC-dependent startup until `GetAccessor`, propagate keystore/TXM startup failures, and cleanly close accessor-owned services.
- Remove the legacy standalone head tracker and keystore-backed direct transmitter.

## Testing

- Added unit coverage for EVM config translation, accessor lifecycle, keystore injection, and multi-node behavior.

## Checklist

- [x] Cross link related PRs (N/A)


integration/pkg/accessors/evm/factory.go
```diff
@@ -64,19 +55,46 @@ func NewFactory(
...
-		errs = append(errs, errors.New(errStr))
-	}
-	return errs
+// NewFactory creates an EVM AccessorFactory from caller-owned reader dependencies.
+//
+// Deprecated: standalone applications should use CreateAccessorFactory, which owns
+// production chainlink-evm clients, head trackers, and transaction managers. This
+// constructor remains available for integrations that inject their own read services;
+// its accessors do not provide a ContractTransmitter.
+func NewFactory(
```
makramkd: **question**: can we just hard-cutover? its EVM-only, so it won't affect anyone else outside of this repo (and PR).

integration/pkg/accessors/evm/chainlink_config.go
```diff
@@ -0,0 +1,169 @@
...
+	// only when those same defaults declare it; changing ChainType alone would
+	// omit required gas and data-availability settings for some rollups.
+	chainTypeName := strings.TrimSpace(info.Type)
+	switch chainTypeName {
+	case "", "anvil", "ethereum", "geth":
+	default:
+		chainType := chaintype.FromSlug(chainTypeName)
+		if !chainType.IsValid() {
+			return nil, fmt.Errorf("unsupported EVM chain type %q", info.Type)
+		}
```
makramkd: Is this section purely for config validation, or is the chain type used somehow by chainlink-evm? The reason I ask is I don't see `chainTypeName` nor `chainType` actually being used downstream, nor `info.Type`.

integration/pkg/accessors/evm/chainlink_config.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	// The standalone database does not contain chainlink-core's evm.heads schema,
+	// so use the production tracker with its supported in-memory saver mode.
+	chain.HeadTracker.PersistenceEnabled = new(false)
+	// Preserve standalone CCV's finalization semantics. Generic chainlink-evm
+	// defaults use a deeper confirmation window, which delays Finality=0 messages
+	// beyond CCV's verifier and E2E execution deadlines on development chains.
+	chain.FinalityDepth = new(uint32(vtypes.ConfirmationDepth))
```
makramkd: Hm I think this needs to be in the EVM CCV configuration as well - this value seemed to be probably for the "simple" head tracker?

integration/pkg/accessors/evm/chainlink_config.go
```diff
@@ -0,0 +1,169 @@
...
+	// These services are not consumers of the standalone accessor. Disabling
+	// them keeps this lifecycle focused on the production HeadTracker and TXM.
+	chain.LogBroadcasterEnabled = new(false)
+	chain.BalanceMonitor.Enabled = new(false)
+	chain.Transactions.Enabled = new(true)
+	chain.Transactions.ForwardersEnabled = new(false)
+	chain.Transactions.TransactionManagerV2.Enabled = new(true)
+	chain.Transactions.TransactionManagerV2.BlockTime = commonconfig.MustNewDuration(defaultTXMBlockTime)
```
makramkd: Same q here, should this be exposed in the CCV config - might be chain-specific (i.e. within EVM chains its varied due to block time).

integration/pkg/accessors/evm/chainlink_config.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	tomlConfig := &evmtoml.EVMConfig{
+		ChainID: sqlChainID,
+		Chain:   chain,
+		Nodes:   nodes,
+	}
+	if err := commonconfig.Validate(tomlConfig); err != nil {
+		return nil, fmt.Errorf("invalid chainlink-evm config for chain %s: %w", info.ChainID, err)
+	}
```
makramkd, line 96: This is odd - do we do this in chainlink-evm as well? Looking at this `Validate` method it does all sorts of things, reflection etc.

integration/pkg/accessors/evm/chainlink_config.go
```diff
@@ -0,0 +1,169 @@
...
+	return evmconfig.NewTOMLChainScopedConfig(tomlConfig), nil
+}
+
+// toChainlinkEVMNode deliberately maps only the endpoint subset owned by CCV.
+// chainlink-evm-only options such as SendOnly, Order, and IsLoadBalancedRPC
+// remain unset and therefore retain their upstream behavior.
+func toChainlinkEVMNode(info Info, index int, configured Node) (*evmtoml.Node, bool, error) {
+	httpURL := firstNonEmpty(configured.InternalHTTPUrl, configured.ExternalHTTPUrl)
```
makramkd, line 107: Unless I'm missing something, I think we need to refactor the Internal/External stuff; I think CTF has this because of docker networks or some such, I don't think it makes sense for nops to be exposed to it. Doesn't have to be in this PR, but could be a follow-up.

integration/pkg/accessors/evm/chainlink_config.go
```diff
@@ -0,0 +1,169 @@
...
+
+	wsURL := firstNonEmpty(configured.InternalWSUrl, configured.ExternalWSUrl)
+	var parsedWS *commonconfig.URL
+	if wsURL != "" {
+		parsedWS, err = commonconfig.ParseURL(wsURL)
+		if err != nil {
+			return nil, false, fmt.Errorf("EVM chain %s node %d has invalid WebSocket RPC URL: %w", info.ChainID, index, err)
+		}
+	}
+
```
makramkd, line 128: Should there be a check here for if both HTTP AND WS are empty, error out? Or is that done elsewhere?

integration/pkg/accessors/evm/chainlink_config.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, value := range values {
+		if value = strings.TrimSpace(value); value != "" {
+			return value
+		}
+	}
+	return ""
+}
+
+func nodeName(info Info, index int) string {
+	base := strings.TrimSpace(info.UniqueChainName)
```
makramkd: UniqueChainName is also odd to have in the prod facing config - I forget what it is in CTF but in the CL node the "name" is an RPC identifier, e.g. "chainstack" or "simplyvc"

integration/pkg/accessors/evm/config.go
```diff
@@ -12,9 +12,12 @@ const (
...
-// NOP-owned settings — RPC endpoints and chain-type tuning — keyed by chain selector. Chain ID
-// and family are derived from the selector, not stored here. This mirrors the Solana and Canton
-// local-config pattern, keeping connection details out of the JD job spec's blockchain_infos.
+// NOP-owned settings — RPC endpoints and optional chain-type assertions — keyed by chain
+// selector. It is intentionally CCV-owned rather than an alias of chainlink-evm's much larger
+// configuration; chainlink_config.go contains the single explicit adapter to that upstream
+// model. Chain ID and family are derived from the selector, not stored here. This mirrors the
+// Solana and Canton local-config pattern, keeping connection details out of the JD job spec's
+// blockchain_infos.
 type Config struct {
```
makramkd, line 22: Can we create a ticket to autogen the docs for this struct using `tools/configdoc`?

integration/pkg/accessors/evm/factory.go
makramkd: Hm yeah, this seems unused from what I can see. Can we just delete it?

integration/pkg/accessors/evm/factory.go
```diff
@@ -64,19 +55,46 @@ func NewFactory(
...
+		executionVisibilityWindow,
+		func(_ context.Context, chainSelector protocol.ChainSelector, _ logger.Logger) (chainRuntime, error) {
+			chainClient, ok := chainClients[chainSelector]
+			if !ok || chainClient == nil {
+				return nil, fmt.Errorf("chain client is not set for chain %d", chainSelector)
+			}
+			return &injectedReaderRuntime{
+				chainClient: chainClient,
+				headTracker: headTrackers[chainSelector],
+			}, nil
```
makramkd: I think this can be deleted too - unless its used for tests and can't be replaced by the other impl. We can maybe use a mock instead there?

integration/pkg/accessors/evm/factory.go
```diff
@@ -85,17 +103,11 @@ func isValidAddress(s string) bool {
...
 
 func (f *factory) GetAccessor(ctx context.Context, chainSelector protocol.ChainSelector) (chainaccess.Accessor, error) {
-	var errs []error
 	if f == nil {
-		errs = append(errs, errors.New("evm accessor factory is nil"))
-	} else {
-		errs = appendErrorIfNil(errs, f.onRampAddresses, "onramp addresses are nil")
-		errs = appendErrorIfNil(errs, f.rmnRemoteAddresses, "rmn remote addresses are nil")
-		errs = appendErrorIfNil(errs, f.headTrackers, "head trackers are nil")
-		errs = appendErrorIfNil(errs, f.chainClients, "chain clients are nil")
```
makramkd, line 95: Sanity checking: is this kind of validation (or similar) done elsewhere in this method?

integration/pkg/accessors/evm/factory.go
```diff
@@ -106,82 +118,108 @@ func (f *factory) GetAccessor(ctx context.Context, chainSelector protocol.ChainS
...
-		if !ok {
+	if hasSourceReaderConfig {
+		headTracker := runtime.HeadTracker()
+		if headTracker == nil {
+			_ = runtime.Close()
 			return nil, fmt.Errorf("head tracker is not set for chain %d", chainSelector)
 		}
 		sr, err := NewEVMSourceReader(
-			chainClient,
+			runtime.ChainClient(),
```
makramkd: Does this need a nil check? (similar to headTracker() above)

integration/pkg/accessors/evm/multi_node_client.go
```diff
@@ -3,122 +3,69 @@ package evm
...
 )
 
-func ptr[T any](t T) *T { return new(t) }
-
-func CreateHealthyMultiNodeClient(ctx context.Context, infos chainaccess.Infos[Info], lggr logger.Logger, chainSelector protocol.ChainSelector) (client.Client, error) {
+// CreateHealthyMultiNodeClient resolves the requested chain and starts its client.
+//
+// Deprecated: use CreateMultiNodeClientFromInfo after resolving the chain info.
+// This wrapper performs no health checks beyond those performed by client.Dial.
+func CreateHealthyMultiNodeClient(
```
makramkd: I don't see any uses of this - can we delete?

integration/pkg/accessors/evm/standalone_chain.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/smartcontractkit/chainlink-evm/pkg/logpoller"
+	"github.com/smartcontractkit/chainlink-evm/pkg/txmgr"
+)
+
+// chainRuntime is the lifecycle boundary owned by one EVM accessor. Keeping it
+// behind this narrow interface makes the accessor wiring testable without
+// replacing chainlink-evm's production implementations.
+type chainRuntime interface {
+	ChainClient() client.Client
+	HeadTracker() heads.Tracker
```
makramkd: Question: better to return `error` in the event these don't exist (I see nil checks above). Otherwise we might miss nil checks somewhere and a panic ensues

integration/pkg/accessors/evm/chainlink_config.go
tt-cll: good catch. The field was only being used as an assertion against the chainlink-evm defaults already selected by chain ID. did not impact downstream

integration/pkg/accessors/evm/factory.go
tt-cll: yeah good call

integration/pkg/accessors/evm/chainlink_config.go
tt-cll: yep agreed

integration/pkg/accessors/evm/chainlink_config.go
tt-cll: yeah it should.

integration/pkg/accessors/evm/chainlink_config.go
tt-cll: agree. opened https://smartcontract-it.atlassian.net/browse/CCIP-12693 to track so we don't keep bloating this PR

integration/pkg/accessors/evm/chainlink_config.go
tt-cll: yep you're right. made this match the CL node impl

integration/pkg/accessors/evm/config.go
tt-cll: https://smartcontract-it.atlassian.net/browse/CCIP-12694

integration/pkg/accessors/evm/factory.go
tt-cll: should be safe to delete

integration/pkg/accessors/evm/multi_node_client.go
tt-cll: yes

integration/pkg/accessors/evm/config.go
```diff
@@ -24,39 +28,42 @@ type Config struct {
...
 	// Nodes are the RPC endpoints used to reach this chain.
 	Nodes []Node `toml:"nodes"`
-	// ChainType selects family-specific EVM client behavior (for example optimismBedrock or
-	// arbitrum). Empty means generic EVM. It is not derivable from the selector, so operators
-	// supply it here.
-	ChainType string `toml:"chain_type,omitempty"`
-	// UniqueChainName is an optional human-readable label used in logs and node names.
-	UniqueChainName string `toml:"unique_chain_name,omitempty"`
+	// FinalityDepth is the number of blocks required before a head is considered final.
+	// Zero uses CCV's default confirmation depth.
```
makramkd: Should zero mean finality tag instead? The default parameter seems like a footgun to me 🤔

tt-cll: another good catch!

build/devenv/env-src-auto-mine.toml
```diff
@@ -6,7 +6,10 @@
   container_name = "blockchain-src"
   chain_id = "1337"
   # When we don't send specific param for mining, anvil defaults to instant mining
-  docker_cmd_params = ["--slots-in-an-epoch", "1", "--disable-code-size-limit"]
+  # Anvil's finalized tag trails the latest block by two epochs. Eight slots
+  # gives this reorg profile a 16-block finality window, matching the tests'
+  # 15-block full-finality assumption while still exercising finality-tag mode.
+  docker_cmd_params = ["--slots-in-an-epoch", "8", "--disable-code-size-limit"]
```
makramkd, line 12: Was the previous config causing flakes? Or is this due to the new components being used?

integration/pkg/accessors/evm/chainlink_config.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	sqlChainID := sqlutil.New(chainID)
+	chain := evmtoml.Defaults(sqlChainID)
+
+	// The standalone database does not contain chainlink-core's evm.heads schema,
+	// so use the production tracker with its supported in-memory saver mode.
+	chain.HeadTracker.PersistenceEnabled = new(false)
+	if info.FinalityDepth == 0 {
+		// FinalityDepth itself must remain positive in chainlink-evm, even in
+		// finality-tag mode, so retain the chain-specific upstream depth.
```
makramkd: This comment seems to suggest that `chain.FinalityDepth` must be set to some positive value, but it isn't set after `chain.FinalityTagEnabled`. Outdated comment?

integration/pkg/accessors/evm/chainlink_config.go
```diff
@@ -0,0 +1,145 @@
...
+	chain.LogBroadcasterEnabled = new(false)
+	chain.BalanceMonitor.Enabled = new(false)
+	chain.Transactions.Enabled = new(true)
+	chain.Transactions.ForwardersEnabled = new(false)
+	chain.Transactions.TransactionManagerV2.Enabled = new(true)
+	blockTime := info.BlockTime
+	if blockTime == 0 {
+		blockTime = defaultTXMBlockTime
+	}
+	chain.Transactions.TransactionManagerV2.BlockTime = commonconfig.MustNewDuration(blockTime)
```
makramkd, line 67: Question about this: if we enable the TXM in the verifier, will it start logging even though its idle? The TXM isn't used in that scenario, and those extra logs would just be useless spam (and potentially misleading).

Two options (possibly more): we can have an `info.TXMEnabled` that we document to set to true for the executor only; we can infer whether to enable it or not based on the app being initialized (unsure if this is easy or not). The latter is better - less knobs is always good.

integration/pkg/accessors/evm/config.go
```diff
@@ -24,39 +28,42 @@ type Config struct {
...
-	// supply it here.
-	ChainType string `toml:"chain_type,omitempty"`
-	// UniqueChainName is an optional human-readable label used in logs and node names.
-	UniqueChainName string `toml:"unique_chain_name,omitempty"`
+	// FinalityDepth is the number of blocks required before a head is considered final.
+	// Zero enables finality-tag mode; a positive value uses confirmation-depth mode.
+	FinalityDepth uint32 `toml:"finality_depth,omitempty"`
+	// BlockTime controls TXM v2's retry cadence. Zero uses the standalone default;
+	// non-zero values must be at least two seconds.
+	BlockTime time.Duration `toml:"block_time,omitempty"`
```
makramkd: Nit: if this is _only_ for TXM block time, we should be verbose and call it so `TXMBlockTime`, to dispel any confusion about it being related to the actual blockchain's block time?

makramkd (commented):
LGTM - just found a couple of more potential gaps in the config part

build/devenv/env-src-auto-mine.toml
tt-cll: it was with testing some of the new components yeah. We weren't actually getting the finality expected during tests

integration/pkg/accessors/evm/chainlink_config.go
tt-cll: yep good catch

integration/pkg/accessors/evm/chainlink_config.go
tt-cll: added a comment to clarify, but the latter option should be how it works today. TXM won't be properly initialized without an off ramp specified, but that key shouldn't exist in the verifier, only the executor

makramkd (approved):

carte7000 (approved):

PR #1297, opened 2026-07-23
feat: verifier & executor traces (CCIP-12516)

bukata-sa:
## Description
  - Add per-message OTel tracing for verifier and executor pipelines.
  - Propagate trace context across queue, heap, retry, task verifier, storage writer, and executor attempt boundaries.
  - Record chain names/selectors, verifier IDs, job IDs, retries, drops, failures, and success events on message spans.
  - Wire `Monitoring.Tracing()` through Beholder, fake, and noop monitoring implementations.

## Checklist


executor/pkg/executor/cl_executor.go
```diff
@@ -156,8 +158,11 @@ func (cle *ChainlinkExecutor) CheckValidMessage(ctx context.Context, message pro
 // 5. Order the CCV data to match the order expected by the receiver contract.
 // 6. Create the aggregated report and transmit it to the chain.
 func (cle *ChainlinkExecutor) HandleMessage(ctx context.Context, message protocol.Message) (shouldRetry bool, err error) {
+	span := oteltrace.SpanFromContext(ctx)
+
 	// Check prerequisites to avoid toctou issues
-	if cle.CheckValidMessage(ctx, message) != nil {
+	if err := cle.CheckValidMessage(ctx, message); err != nil {
+		span.AddEvent("invalid_message_skipped")
```
makramkd: Can we make these constants (can be in this file) starting from now, so that we can potentially build tooling on top (e.g. debugging tools that check for these kinds of events)

executor/pkg/executor/cl_executor.go
```diff
@@ -179,20 +187,26 @@ func (cle *ChainlinkExecutor) HandleMessage(ctx context.Context, message protoco
...
 
 	verifierResults, verifierQuorum, err := cle.getVerifierResultsAndQuorum(ctx, message, messageID)
 	if err != nil {
 		cle.lggr.Warnw("delaying execution due to failed request for verifier results and quorum", protocol.LogKeyMessageID, messageID)
+		span.RecordError(err)
+		span.AddEvent("delayed_verifier_results_error")
 		return true, err
 	}
 	if len(verifierResults) == 0 {
+		span.AddEvent("delayed_no_verifier_results")
```
makramkd: Do we only record an error if we have an `err` variable to use? This block ends up returning an error, but we don't RecordError

executor/pkg/executor/cl_executor.go
```diff
@@ -274,6 +299,10 @@ func (cle *ChainlinkExecutor) HandleMessage(ctx context.Context, message protoco
 	// Record the message execution latency.
 	cle.monitoring.Metrics().RecordMessageExecutionLatency(ctx, time.Since(start), destinationChain)
 
+	span.AddEvent("message_transmitted", oteltrace.WithAttributes(
+		attribute.Int64("latest_ccv_timestamp", latestCCVTimestamp),
```
makramkd: Attributes should be constants too - if they are usable by debugging tools

executor/executor_coordinator.go
```diff
@@ -39,22 +43,13 @@ type Coordinator struct {
...
-	executor Executor,
-	messageSubscriber MessageSubscriber,
-	leaderElector LeaderElector,
-	monitoring Monitoring,
-	expiryDuration time.Duration,
-	timeProvider common.TimeProvider,
-	workerCount int,
-	dataNotReadyRetryInterval time.Duration,
-) (*Coordinator, error) {
+func NewCoordinator(lggr logger.Logger, executorID string, executor Executor, messageSubscriber MessageSubscriber, leaderElector LeaderElector, monitoring Monitoring, expiryDuration time.Duration, timeProvider common.TimeProvider, workerCount int, dataNotReadyRetryInterval time.Duration) (*Coordinator, error) {
```
makramkd: This is a breaking change, so I recommend a PR title of "feat!" and not "feat" (notice the "!")

executor/executor_coordinator.go
```diff
@@ -212,15 +207,32 @@ func (ec *Coordinator) runStorageStream(ctx context.Context) {
 				"readyTimestamp", readyTimestamp,
 			)
 
+			// Discovery span: the per-message root for this instance's
+			// processing. It stays open across delay/retry cycles (via
+			// TraceContext propagated through the heap) and is only ended at
+			// a terminal outcome (expired, executed, or permanently failed)
+			// in processPayload.
+			discCtx, discSpan := ec.monitoring.Tracing().StartMessageSpan(
+				ctx, "executor.message@"+ec.executorID, id,
```
makramkd: This `"executor.message@"+ec.executorID` should be a free function IMO

executor/executor_coordinator.go
```diff
@@ -212,15 +207,32 @@ func (ec *Coordinator) runStorageStream(ctx context.Context) {
...
+				attribute.String("dest_chain_selector", msg.DestChainSelector.String()),
+				attribute.String("source_chain_selector", msg.SourceChainSelector.String()),
```
makramkd: These should be (common?) constants, so that we don't have attribute drift across the codebase.

verifier/migrations/postgres/00008_add_verifier_results_traceparent.sql
```diff
@@ -0,0 +1,5 @@
+-- +goose Up
+ALTER TABLE verifier_node_results ADD COLUMN traceparent TEXT NOT NULL DEFAULT '';
```
makramkd, line 2: Hm why would we need to persist the trace parent? In the event of crashes?

verifier/pkg/sourcereader/service.go
```diff
@@ -432,6 +461,19 @@ func (r *Service) addToPendingQueueHandleReorg(tasks []verifier.VerificationTask
 		existingBlock := new(big.Int).SetUint64(existing.BlockNumber)
 		if existingBlock.Cmp(fromBlock) >= 0 && (toBlock == nil || existingBlock.Cmp(toBlock) <= 0) {
 			if _, exists := tasksMap[msgID]; !exists {
+				if existing.TraceContext != nil {
```
makramkd: This `TraceContext != nil` check seems to be a ticking time-bomb for an eventual nil-pointer dereference panic; is there something we can do to avoid the nil checks before use on every call site?

verifier/migrations/postgres/00008_add_verifier_results_traceparent.sql
makramkd: Also, I think, but am not sure, that this migration will also need to be applied on the chainlink node migrations.

makramkd (commented):
General thoughts:

1. The additional tracing code is quite invasive and in some places delicate (TraceContext != nil checks, for example). I'm wondering if there is a small library/pattern that we can create/adopt to make the usage of the tracing API easier and less error-prone.
2. The way Go code is written it is quite easy to accidentally miss a `span.End()` (or similar) call before a `return`. Can we lean into `defer` more to ensure we don't miss these kinds of calls?
3. Placed comments on the code, but IMO we need to better organize the attribute labels (e.g. "source_chain_selector") and any other identifiers as constants in a probably common package (common/monitoring/tracing seems fine) so that all the apps (verifier/aggregator/indexer/executor) all use the same attribute labels.

verifier/migrations/postgres/00008_add_verifier_results_traceparent.sql
bukata-sa: verifier_node_results doesn't exist in core, Committee Verifiers push to aggregator
This table exists only in standalone, for token_verifier

common/monitoring/tracing/keys.go
```diff
@@ -0,0 +1,23 @@
+package tracing
+
+// Span attribute keys used across the ccip.
```
makramkd, line 3:
```suggestion
// Span attribute keys used across the ccip apps.
```

verifier/pkg/jobqueue/testdata/explain_cleanup.txt
makramkd, line 1: These explain*.txt files seem like just noise, I don't think they're checked in CI? Can we delete them in a follow-up, or enforce the query plan somehow (e.g. index scan always) in CI?

makramkd (commented):
Couple more comments - I guess the tracing code is necessarily verbose. We may be able to do some small cleanups but I don't expect too much.

One question remaining: is there a way we can test the traces in _unit tests_? At least, that way we can have quick assertions for changes that happen to the tracing code.

PR #1301, opened 2026-07-27
feat: Add KMS support

AmrMohamedRezk:
## Description
Add KMS signer support

## Testing
- Unit tests
- Tested against live environment.

## Checklist


bootstrap/keys/kms.go
```diff
@@ -0,0 +1,154 @@
...
+	return newKMSKeystore(ctx, inner, nameToID)
+}
+
+// newKMSKeystore wraps an existing Reader+Signer with logical-name translation and verifies that
+// every mapped key exists.
+func newKMSKeystore(ctx context.Context, inner signerReader, nameToID map[string]string) (*KMSKeystore, error) {
+	idToName := make(map[string]string, len(nameToID))
+	for name, id := range nameToID {
+		if existing, ok := idToName[id]; ok {
+			return nil, fmt.Errorf("KMS key ID %q is mapped to both %q and %q: each key ID must map to exactly one logical name", id, existing, name)
```
makramkd: Question: are the key IDs sensitive info?

bootstrap/keys/kms.go
```diff
@@ -0,0 +1,154 @@
...
+			if friendly, ok := k.idToName[resp.Keys[i].KeyInfo.Name]; ok {
+				resp.Keys[i].KeyInfo.Name = friendly
+			}
+		}
+		return resp, nil
+	}
+
+	// The configured name->ID map is the allowlist: reject any name that is not in it, rather than
+	// forwarding it to KMS. Otherwise a caller of the (caller-controlled) info server getkeys endpoint
+	// could probe any KMS key the bootstrapper's IAM role can reach, not just the intended set.
```
makramkd, line 116: I believe that it would be worth making this super clear in the documentation of the KMS config (creating super restrictive IAM roles for specific keys)

bootstrap/bootstrap.go
```diff
@@ -500,32 +504,47 @@ func (b *Bootstrapper) startWithJDLifecycle(ctx context.Context) error {
...
-			return fmt.Errorf("failed to connect to bootstrapper database: %w", err)
+			return fmt.Errorf("failed to initialize KMS keystore: %w", err)
 		}
-		keyStore, csaSigner, err = initializeKeystore(ctx, b.lggr, db, ksPassword, b.keys)
-		if err != nil {
-			return fmt.Errorf("failed to initialize keystore: %w", err)
+	default:
+		// Postgres backend: A keystore is initialized only when both the DB URL and keystore password are configured.
+		dbURL := strings.TrimSpace(b.config.DB.URL)
+		ksPassword := strings.TrimSpace(b.config.Keystore.Password)
```
makramkd, line 529: I'm not sure if we should trim spaces on a password - I guess technically speaking a password that looks like `"   this-is my password   "` is valid?

Unless the keystore library we're using doesn't allow untrimmed strings as passwords. In that case it should be a validation on the `Keystore` config

bootstrap/bootstrap.go
```diff
@@ -500,32 +504,47 @@ func (b *Bootstrapper) startWithJDLifecycle(ctx context.Context) error {
...
+		switch {
+		case dbURL != "" && ksPassword != "":
```
makramkd, line 531: Oh I see, the trimming is just to do the emptiness check... that makes sense. Does a password of all spaces make sense to reject at the validation stage then (in KeystoreConfig.validate())?

bootstrap/bootstrap.go
```diff
@@ -500,32 +504,47 @@ func (b *Bootstrapper) startWithJDLifecycle(ctx context.Context) error {
...
+			if err != nil {
+				return fmt.Errorf("failed to initialize keystore: %w", err)
+			}
+		case dbURL != "" || ksPassword != "":
+			// Exactly one of [db]/[keystore] is set — almost certainly a misconfiguration. Warn loudly:
+			// running keystore-less here would surface later as a confusing nil-keystore failure in a
+			// signing service.
+			b.lggr.Warnw("local mode: bootstrap config sets only one of [db].url / [keystore].password; "+
+				"both are required to initialize the keystore, so the service will run keystore-less",
+				"hasDBURL", dbURL != "", "hasKeystorePassword", ksPassword != "")
```
makramkd, line 546: Looking at this now, I'm also wondering if this should be an overall config validation failure (fails fast) vs. a warning log and some error in the service startup.

bootstrap/config.go
```diff
@@ -33,15 +33,78 @@ func (c *JDConfig) validate() error {
...
+		return err
+	}
+	if backend == KeystoreBackendKMS {
+		// The bootstrapper always needs an Ed25519 CSA key (auto-injected in NewBootstrapper and used
+		// by keys.NewCSASigner), so ed25519_key_id is required for the KMS backend. ecdsa_key_id is
+		// service-specific — only services that declare an ECDSA signing key need it — so its presence
+		// is enforced per-declared-key in buildKMSNameMap rather than statically here.
+		if c.KMS.Ed25519KeyID == "" {
+			return fmt.Errorf("field 'ed25519_key_id' is required when backend is 'kms' (the CSA key is Ed25519)")
+		}
```
makramkd: I don't believe this is technically true - we should be able to use KMS as the signer w/out using JD (see the "local app config" code path). This would probably be the case for all 3rd party usages of KMS (JD isn't public).

tools/configdoc/registry/registry.go
```diff
@@ -274,10 +274,21 @@ func bootstrapConfigInstance() any {
 // bootstrapSecretsInstance builds the documented bootstrap secrets file
 // (bootstrap.Secrets, the half devenv marshals into secrets.toml, loaded via
 // BOOTSTRAPPER_SECRETS_PATH) with obviously-fake placeholder credentials.
+// The keystore section shows the default postgres backend; the KMS backend is
+// documented via struct comments on KeystoreConfig and KMSKeystoreConfig.
```
makramkd: Stale (I believe) comment

build/devenv/services/common.go
```diff
@@ -25,6 +25,32 @@ const (
 	AppPathInsideContainer = "/app"
 )
 
+// awsCredentialEnvVars are the standard AWS SDK environment variables that carry credentials and
+// region. Forwarding these from the host lets a container reach AWS (e.g. KMS) via the default
+// credential chain without a mounted profile. Session-token creds (SSO/STS) are short-lived, so a
+// forwarded set expires with the host session.
+var awsCredentialEnvVars = []string{
```
makramkd, line 32: I think this is fine, but I feel like it should be better documented (this is a private variable - won't show up even in pkg.go.dev generated docs).

Maybe a small note in the README or the associated changelog, on how to test w/ KMS in devenv?

bootstrap/config.go
```diff
@@ -33,15 +33,78 @@ func (c *JDConfig) validate() error {
...
+	// KeystoreBackendKMS stores keys in AWS KMS; private key material never leaves KMS.
+	KeystoreBackendKMS KeystoreBackend = "kms"
+)
+
+// KMSKeystoreConfig configures the AWS KMS keystore backend.
+// Field names match the pricer's KMSConfig for future consolidation.
+type KMSKeystoreConfig struct {
+	// Profile is the AWS shared-config profile name (for local development).
+	// If empty, uses the default credential chain (IRSA, EC2 instance profiles, etc.).
+	Profile string `toml:"profile,omitempty"`
```
makramkd: Is this for local development _only_ or would this be used in production deployments?

bootstrap/keys/kms.go
AmrMohamedRezk: They aren't sensitive. But they do include region and AWS account. Probably better to not log them. I will adjust the logging.

bootstrap/bootstrap.go
AmrMohamedRezk: Yeah that was the existing behavior I haven't changed it and I think it makes sense.

bootstrap/bootstrap.go
AmrMohamedRezk: I think it should fail, i guess if no KMS and no DB that's not a healthy state we want to startup with right ?

bootstrap/config.go
AmrMohamedRezk: profile is local only. Will add a clarifying comment

bootstrap/bootstrap.go
makramkd: Oh nvm, I guess there might be a usecase here - token verifier needs a DB but doesn't need a keystore 🤔

bootstrap/config.go
```diff
@@ -33,15 +33,84 @@ func (c *JDConfig) validate() error {
...
+	// Password is the password to the keystore. Required when backend is "postgres" (default).
+	Password string `toml:"password,omitempty"`
+	// KMS configures the AWS KMS backend. Required when backend is "kms".
+	//
+	// IAM: scope the bootstrapper's IAM role as tightly as possible. Grant only kms:Sign,
+	// kms:GetPublicKey, and kms:DescribeKey, and restrict Resource to exactly the key ARNs configured
+	// below (ecdsa_key_id / ed25519_key_id). Do NOT grant kms:ListKeys or a wildcard Resource ("*").
+	// The service limits operations to the configured keys, but a per-key least-privilege IAM
+	// policy is the primary safeguard: it guarantees these credentials cannot read or sign with any
+	// other KMS key in the account.
```
makramkd, line 74: 💯

makramkd (approved):

carte7000 (approved):

PR #1309, opened 2026-07-29
feat(evm): standalone migration testing

tt-cll:
## Checklist


docs/migration/cl-to-standalone.md
```diff
@@ -0,0 +1,224 @@
+# Migrating an EVM node operator from CL mode to standalone
+
+A node operator running CCV as jobs on a Chainlink node moves to the standalone verifier and
+executor without touching a contract, without moving funds, and without a second entry in the Job
+Distributor. This document is the procedure. It assumes EVM; Solana and Canton were standalone from
+the start and have nothing to migrate.
```
makramkd:
```suggestion
Distributor. This document is the procedure. It assumes EVM; other families are standalone from
the start and have nothing to migrate.
```

docs/migration/cl-to-standalone.md
```diff
@@ -0,0 +1,224 @@
...
+
+## What has to carry over, and why
+
+Three pieces of the Chainlink node's identity are visible outside it.
+
+The **OCR2 EVM key bundle's onchain signing key** is the address in the `CommitteeVerifier` signer
+set. The node publishes it to JD as `OnchainSigningAddress`, and `ApplyVerifierConfig` reads it back
+from there when it configures the contract. Generate a new one and the committee's signer set has to
+be updated on every chain, which needs a config transaction per chain and coordination with the
+other operators in the committee.
```
makramkd:
```suggestion
from there when it configures the contract. If you generate a new one, the committee's signer set has to
be updated on every chain, which needs a config transaction per chain and coordination with the
other operators in the committee.
```

I'm not sure of the intended audience of the doc - if its node operators, its probably not necessary to have this sentence.

docs/migration/cl-to-standalone.md
```diff
@@ -0,0 +1,224 @@
...
+
+The **EVM account key** is the transmitter the executor submits from. It is the address JD records
+as `AccountAddr`, and it holds the gas the operator funded it with. Generate a new one and the
+operator moves funds.
```
makramkd:
```suggestion
as `AccountAddr`, and it holds the gas the operator funded it with. If you generate a new one, the
operator has to move funds to the new key.
```

docs/migration/evm-cl-to-standalone.md
```diff
@@ -0,0 +1,224 @@
...
+
+The **EVM account key** is the transmitter the executor submits from. It is the address JD records
+as `AccountAddr`, and it holds the gas the operator funded it with. Generate a new one and the
+operator moves funds.
+
+The **CSA key** authenticates the node to JD. It carries no on-chain meaning, so it is not exported.
+Instead the JD node record is repointed at the standalone verifier's own CSA key, which preserves
+the node ID, its name, and its job history while the key stays on the machine that generated it.
+This is also why the Chainlink node must be stopped before the standalone verifier starts: JD
+identifies a node by the key it authenticates with, and one record cannot have two owners.
```
makramkd, line 29: Question: it must only be stopped once we intend to send jobs to the standalone verifier and executor nodes, right?

docs/migration/evm-cl-to-standalone.md
```diff
@@ -0,0 +1,224 @@
...
+
+```sh
+chainlink jobs list
+```
+
+If the node runs more than one verifier job, stop and raise it before going further. That happens
+when the committee's job specs are still generated per aggregator, which is a property of how
+Chainlink Labs runs the aggregators and generates job proposals — nothing a node operator
+configures or can change. It has to be resolved on that side first, by consolidating the committee's
+verifier jobs into one that writes to every aggregator, and only then does this procedure apply.
```
makramkd, line 63: Aside: any idea if we've done this migration yet in prod (single job, multiple aggregators)

docs/migration/cl-to-standalone.md
```diff
@@ -0,0 +1,224 @@
...
+process's EVM config path and it is read directly: the `[[EVM]]` and `[[EVM.Nodes]]` sections are
+translated at startup, chain IDs are resolved to chain selectors, and every other section in the
+file is ignored.
+
+Finality carries over as the node had it. The node's own chain defaults are applied before
+translating, so a chain you never configured explicitly keeps the behavior it had rather than
+quietly moving onto finality tags.
+
+Two things a Chainlink node can express have no standalone equivalent and are dropped: send-only
+nodes, and per-node `Order` and `HTTPURLExtraWrite`. Each one is logged at startup saying exactly
```
makramkd: Aside: why did we not add `Order`? Not applicable?

docs/migration/cl-to-standalone.md
```diff
@@ -0,0 +1,224 @@
...
+
+Two paths and one check. You do not say which keystore key the file becomes, because each process
+has exactly one it can import into, and you do not say which export it is, because that is read from
+the file.
+
+`expected_id` is the address from step 1: the signing address for the verifier, the account address
+for the executor. It is optional and safe to skip only if you are migrating exactly one node.
+Mounting the wrong node's export otherwise brings up a process that signs with another operator's
+key, which produces verification results the committee rejects with nothing in the logs pointing at
+the cause. Set it.
```
makramkd: `expected_id` is a good sanity check - I don't think it should be optional, since it seems pretty straightforward to set. It would also trim some of the documentation here. Giving people optional stuff usually means they skip them out of laziness (me included).

docs/migration/evm-cl-to-standalone.md
makramkd, line 1: Nit on doc name: `evm-cl-to-standalone.md` to make it clear this is EVM targeted.

docs/migration/cl-to-standalone.md
```diff
@@ -0,0 +1,224 @@
...
+The port is the one you set as `listen_port` in the bootstrap config's `[server]` section; 9988 is
+the value used throughout this document.
+
+The same endpoint returns the imported signing key, but as a public key rather than an address —
+deriving the address from it is a keccak hash of the uncompressed key, not something to do by eye.
+Rely on `expected_id` instead: the process refuses to start if the key it imported, or the key
+already in its keystore, is not the address you pinned. Reaching a healthy state is the
+confirmation.
+
+## Step 6: hand the JD record to the verifier, and register the executor
```
makramkd: Steps are a mix of node operator actions and CCIP engineer actions, would be nice to make it very clear which is for which.

bootstrap/config.go
```diff
@@ -150,6 +150,48 @@ type ChainRegistration struct {
 	ID string `toml:"id"`
 }
 
+// KeyImport declares a Chainlink node key to adopt into the keystore instead of generating a fresh
+// one. It exists for node operators moving from CL mode to standalone: the onchain signing key is
+// registered in the CommitteeVerifier signer set and the transmitter key holds gas funds, so both
+// have to carry over. The import runs only when the key is absent, which makes it a no-op on every
```
makramkd:
```suggestion
// one. It exists for node operators moving from CL mode to standalone. The import runs only when the key is absent, which makes it a no-op on every
```

Removing an unnecessary detail not related to the import itself

bootstrap/config.go
```diff
@@ -150,6 +150,48 @@ type ChainRegistration struct {
...
+// The export format is read from the file for the same reason.
+type KeyImport struct {
+	// Path is the mounted path of the key file exported from the Chainlink node.
+	Path string `toml:"path"`
+	// PasswordPath is the mounted path of a file holding the password used at export time.
+	PasswordPath string `toml:"password_path"`
+	// ExpectedID optionally pins the address the export must carry. Set it when migrating more than
+	// one node, so mounting the wrong node's file fails at startup instead of bringing up a process
+	// that signs as somebody else.
+	ExpectedID string `toml:"expected_id,omitempty"`
```
makramkd: Lets make this required - optional usually means will never get used 😛 

I would also suggest removing the "migrate more than one node" reasoning/framing - it should simply be framed as required in order to ensure the key import is successful.

bootstrap/README.md
```diff
@@ -135,6 +135,40 @@ type = "EVM"   # chain family — EVM, SOLANA, APTOS, STELLAR, CANTON, STARKNET,
 id   = "1"     # chain ID (e.g. EVM chain ID, Solana cluster name)
 ```
 
+## Adopting a key from a Chainlink node
+
+A node operator moving off CL mode has keys that must survive the move: the onchain signing key
```
makramkd:
```suggestion
A node operator moving off CL mode has keys that must survive the move: e.g. the onchain signing key
```

bootstrap/README.md
```diff
@@ -135,6 +135,40 @@ type = "EVM"   # chain family — EVM, SOLANA, APTOS, STELLAR, CANTON, STARKNET,
...
+# The password it was exported under, in its own file so this config carries no credentials.
+password_path = "/etc/ccv/migration/export-password.txt"
+# Optional, and worth setting when migrating more than one node: the address the export must carry.
+expected_id   = "0x1234...abcd"
+```
+
+The `ccv migrate` commands in the verifier and executor images produce all of this: `ccv migrate
+export` exports the key from the node, verifies it, and writes this block with `expected_id`
+already filled in; `ccv migrate inspect` prints the identity a mounted file carries, to confirm the
+right file before boot.
```
makramkd, line 158: Are these commands implemented in this PR, or is this an intended follow-up? Or, might be stale docs.

bootstrap/config.go
```diff
@@ -203,6 +245,11 @@ type NonSecretConfig struct {
 	// Optional: if empty, no signing key sync is performed.
 	Chains []ChainRegistration `toml:"chains"`
 
+	// KeyImport adopts a Chainlink node key into the keystore on first boot instead of generating
+	// one. Optional: when absent every declared key is generated, which is right for a new
+	// deployment. It is set when migrating an operator off CL mode.
+	KeyImport *KeyImport `toml:"key_import"`
```
makramkd, line 254: I was wondering whether this should be `[]KeyImport` instead, reasoning being that _maybe_ in the future we'd want to import multiple keys - but I'd rather this be a very directed use case (in order to migrate CL nodes to standalone) and deprecate in the future, _until_ we get some general "key import" feature requirement.

bootstrap/keys/import.go
makramkd, line 1: Nit: wondering whether to have a dedicated subpackage inside keys for this: `keys/evmmigration`, since its tailored to specifically migrating EVM CL nodes? Ideally, when all nodes are migrated we can deprecate and delete the code.

bootstrap/bootstrap.go
```diff
@@ -810,15 +822,19 @@ func initializeKeystore(ctx context.Context, lggr logger.Logger, ksCfg KeystoreC
 			return nil, nil, fmt.Errorf("failed to load keystore: %w", err)
 		}
 		for _, k := range requiredKeys {
-			if err := keys.EnsureKey(ctx, lggr, ks, k.name, k.purpose, k.keyType); err != nil {
+			if importKeyName != "" && k.name == importKeyName {
+				if err := keys.EnsureImportedKey(ctx, lggr, ks, k.name, k.purpose, k.keyType, importSpec); err != nil {
+					return nil, nil, fmt.Errorf("failed to import key %q (purpose=%q, type=%v): %w", k.name, k.purpose, k.keyType, err)
+				}
```
makramkd, line 828: Question: what would happen if there already exists a key w/ this name, would it get overwritten or would this error out?

Practically, how this could happen: a nop launches the standalone verifier/executor, and w/out any jobs/etc. configured and the key_import not present, I believe the keystore would be initialized w/ the keys declared in the `bootstrap.Run` options, if I'm not mistaken.

bootstrap/config.go
tt-cll: agree. Let's keep it singular for now and then expand to multiple if/when required

bootstrap/README.md
tt-cll: added them as part of the stack

